PR #1, opened 2022-11-29
Fix security config

the-tatanka:


nicoprow:
Closing old pull request as suggested changes have been implemented already

PR #11, opened 2023-01-26
Feat/pool changelog filters

martinfkaeser:


bpdm-pool/pom.xml
```diff
@@ -130,6 +130,11 @@
             <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-kotlin</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate.orm</groupId>
+            <artifactId>hibernate-jpamodelgen</artifactId>
+            <version>${hibernate.version}</version>
+        </dependency>
```
nicoprow, line 137: Sorry guys, I misunderstood where this functionality comes from. Direct Hibernate dependencies are problematic and if we have a good alternative we should try to avoid them. Hibernate dependencies run under the LGPL 2.1 license. We should avoid such licenses whenever possible. So if you can implement the functionality without this dependency this would be preferable.

nicoprow (changes requested):
As far as I have seen we can easily avoid the Hibernate dependency with only a little inconvenience. Please remove the dependency and implement the alternative with reflection and type casting.

Also the contribution should increase the application version to a new snapshot version. Since they introduce a breaking change on the business partner endpoint use version: 4.0.0-SNAPSHOT

In order to successfully submit a pull request here you need to sign the Eclipse Committers Agreement. For this, you need to create an account at Eclipse [eclipse](https://www.eclipse.org/), sign the agreement there and link your Github account.

bpdm-pool/pom.xml
martinfkaeser: I removed the dependency and changed the implementation to field reflection.

nicoprow (approved):

nicoprow:
The changes to the Changelog endpoint are breaking changes and belong to the 4.0 version. Please make a Pull Request to the appropriate release candidate branch: rc/4.0.0

PR #37, opened 2023-02-09
Refactor/rename cdq saas

martinfkaeser:


nicoprow (changes requested):
Please revert the changes to the charts directories (README and values). Due to our release process we can currently not change the chart and the underlying app in a breaking change at the same time.

PR #49, opened 2023-02-16
Refactor/rename cdq saas - BREAKING API CHANGES

martinfkaeser:


nicoprow (changes requested):
Since the SyncRecord type is changed we need a database migration of syncrecords from type CDQ_Import to SAAS_IMPORT.
Migration scripts are added to the db.migration folder in the resources of the pool and follow the convention: V\_*major*\_*minor*\_*patch*\_*increment*\_\_*description*.sql

nicoprow (approved):

PR #53, opened 2023-02-22
Feat/generic json deserializer

martinfkaeser:


nicoprow (changes requested):
Looks good but please add a small line in the Changelog for fixes that result in the introduction of the generic deserializer

nicoprow (approved):

PR #62, opened 2023-03-02
Fix/gate parent relations errors

martinfkaeser:


nicoprow (changes requested):
Please add a Changelog entry to tell users about the fix

nicoprow (approved):

PR #93, opened 2023-03-27
Test charts with locally build image

nicoprow:
Changes the helm test workflow to first build their own BPDM images on pull request and then run helm tests on those on locally build images.

This enables us to keep the Helm Charts up-to-date with the app code without needed to publish a new image first.


nicoprow:
This feature would require is to keep the charts up to date with the newest application version. This leads to higher dependencies between app development and chart development and could lead to pull request "bloat". Thus we keep the release cycles between app and chart separate for now.

PR #94, opened 2023-03-30
Feat - Persistence of shared business partners (Address)

alexsilva-CGI:
When a sharing member shares a business partner via the Gate API, the information can be persisted in the database. If the information is already present in the database, then the entries are overwritten and updated. First implementation for the Addresses logic.


nicoprow (changes requested):
There are a lot of commits in the pull request and they often contain a lot of generic descriptions. I would aggressively merge the commits to 1 or 2 commits, create descriptive header descriptions and add details as needed in the commit message body

nicoprow:
New gate persistence branch for addresses will be used for merge

PR #103, opened 2023-04-19
Feat/legal entity persistence logic

cezaralexandremorais:
legal entity persistence logic


nicoprow:
Another branch handling the gate persistence logic for legal entities will be opened

PR #116, opened 2023-05-03
refactor(Maven): Cleanup dependencies in POMs

martinfkaeser:


nicoprow (changes requested):
Squash the two commits with the same description

PR #120, opened 2023-05-03
BPDM bridge dummy: Setup Spring application

martinfkaeser:


bpdm-bridge-dummy/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM maven:3.8.7-eclipse-temurin-17 AS build
+COPY . /home/app
+WORKDIR /home/app
+RUN mvn -B -U clean package -pl bpdm-bridge-dummy -am -DskipTests
+
+FROM eclipse-temurin:17-jre-alpine
+COPY --from=build /home/app/bpdm-gate/target/bpdm-bridge-dummy.jar /usr/local/lib/bpdm/app.jar
+RUN apk update && apk upgrade libssl3 libcrypto3
+RUN addgroup -S bpdm && adduser -S bpdm -G bpdm
```
nicoprow, line 9: Just hcnaged how we add groups and users in the Dockerfile to better match with the default Helm Chart security context. Have a look at the Dockerfiles for pool and gate to see the new way on how its done

nicoprow (changes requested):

bpdm-bridge-dummy/Dockerfile
martinfkaeser: Fixed it

nicoprow (approved):

PR #131, opened 2023-05-09
Feat/field quality rules

rainer-exxcellent:
Pool: Optional, mandatory address fields of business partners
 create new endpoint to access via country ISO Code

based on branche feat/adapt_datamodel


nicoprow:
Updates issue #165

nicoprow (approved):

nicoprow:
The huge lines of code are due to a migration file filling up the database with default content. I don't think Eclipse IP issues is necessary here.

PR #138, opened 2023-05-10
Bridge dummy: simple sync for L/S/A

martinfkaeser:


nicoprow (approved):
Ok, works as a service you as a basis to enhance. But ToDos and open points have been made clear. Basic sync between gate and pool works as well.

Issue #152, opened 2023-05-12
Dummy Bridge Uses Old Datamodel


nicoprow:
Currently the BPDM project does not compile, due to the Dummy Bridge module failing. The Bridge still works against the old API client which leads to errors.

Solution:
Update how the Bridge uses the Gate and Pool API clients to the newest version of the datamodel.

martinfkaeser:
Fixed with https://github.com/eclipse-tractusx/bpdm/pull/166/files

PR #166, opened 2023-05-16
fix(bridge): Adjust to data model changes and add module to POM

martinfkaeser:


martinfkaeser:
Fixes https://github.com/eclipse-tractusx/bpdm/issues/152

rainer-exxcellent:
Approved: data model changes and module are in the right place, gate and pool compile and start

Issue #139, opened 2023-05-10
Create Helm Umbrella Chart


nicoprow:
Currently we have two separate Charts for the BPDM Gate and BPDM Pool services. In order to ease the deployment of the two services we should  offer an umbrella Helm Chart.

The umbrella Helm chart should use the BPDM Gate and BPDM Pool chart as child dependencies, which can be enabled and disabled (enabled by default). On default, the BPDM Gate chart should connect to the deployed BPDM Pool instance by default. 

For the case when the pool is disabled, the BPDM Gate Values can be overriden to connect to an existing pool instance.

fabiodmota:
Done and waiting on review on commit 
https://github.com/catenax-ng/tx-bpdm/commit/7e340f1fd8618a51aaddc2039baedbd940ba7b5e

nicoprow:
Chart collides with old name of Pool Chart and therefore chart releaser complains about version conflicts. We will introduce a major version bump when dealing with issue #151 to solve this

PR #167, opened 2023-05-16
feat(Gate-Pool): Deploy all helm charts with umbrella structure

fabiodmota:
Change Chart Structure for umbrella helm charts
issue https://github.com/eclipse-tractusx/bpdm/issues/139


charts/bpdm/values.yaml
```diff
@@ -0,0 +1,81 @@
...
+    repository: catenax-ng/tx-bpdm/gate
+    pullPolicy: Always
+    tag: ""
+  # ... more configuration for bpdm-gate if need...
+  applicationConfig:
+    bpdm:
+      pool:
+        base-url: ""
+
+bpdm-pool:
```
nicoprow, line 15: I know due to limitations of helm currently child of child dependencies are ignored. But I feel for being more clear here we should still disable postgres and opensearch dependencies here by setting enabled: false

charts/bpdm/values.yaml
```diff
@@ -0,0 +1,81 @@
+bpdm-gate:
```
nicoprow, line 1: I know due to limitations of helm currently child of child dependencies are ignored. But I feel for being more clear here we should still disable postgres dependency here by setting enabled: false

charts/bpdm/values.yaml
```diff
@@ -0,0 +1,81 @@
...
+  enabled: true
+  replicaCount: 1
+  image:
+    registry: ghcr.io
+    repository: catenax-ng/tx-bpdm/pool
+    pullPolicy: Always
+    tag: ""
+  opensearch:
+    masterService: ""
+    fullnameOverride: "bpdm-pool-opensearch"
```
nicoprow, line 25: Is opensearch naming needed here? I feel since the dependency is not created should be not needed, right?

nicoprow (changes requested):
Some minor clarifications needed

charts/bpdm/values.yaml
fabiodmota: done

charts/bpdm/values.yaml
fabiodmota: for now as it stands we have a need to have it like this , we will take a look further on this with the help of dev sec team

nicoprow (approved):

PR #184, opened 2023-05-23
Feat/persistence query input legalentities

cezaralexandremorais:
Changed the fetch of legalentities to call persisted data instead of calling SaaS


bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/LegalEntityService.kt
```diff
@@ -58,7 +63,7 @@ class LegalEntityService(
         legalEntities.forEach { legalEntity ->
             changelogRepository.save(ChangelogEntry(legalEntity.externalId, LsaType.LegalEntity))
         }
-        legalEntityPersistenceService.persistLegalEntitiesBP(legalEntities);
+        legalEntityPersistenceService.persistLegalEntitiesBP(legalEntities)
     }
 
     fun getLegalEntityByExternalId(externalId: String): LegalEntityGateInputResponse {
```
nicoprow, line 69: This method fetches from SaaS still. Should also use the database to fetch the business partner

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/LegalEntityService.kt
```diff
@@ -70,16 +75,16 @@ class LegalEntityService(
...
+    fun getLegalEntities(page: Int, size: Int, externalIds: Collection<String>? = null): PageLegalEntityResponse<LegalEntityGateInputResponse> {
 
-        val validEntries = toValidLegalEntities(partnerCollection.values)
-
-        return PageStartAfterResponse(
-            total = partnerCollection.total,
-            nextStartAfter = partnerCollection.nextStartAfter,
-            content = validEntries,
-            invalidEntries = partnerCollection.values.size - validEntries.size
+        val legalEntitiesPage = legalEntityRepository.findByExternalIdIn(externalIds, PageRequest.of(page, size))
```
nicoprow, line 80: When I execute an endpoint request I get an error here, since the repo method does not expect null as a valid value. Should be empty instead

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/LegalEntityService.kt
```diff
@@ -138,6 +143,21 @@ class LegalEntityService(
             externalId = externalId
         )
 
+    private fun toValidLegalEntities(legalEntityPage: Page<LegalEntity>): List<LegalEntityGateInputResponse> {
+
+        return legalEntityPage.content.map { legalEntity ->
```
nicoprow, line 148: Is this method really needed to create a LegalEntityGateInputResponse object? I would expect a mapping method for LegalEntity to LegalEntityGateInputResponse in the ResponseMappings. Then just use this mapping directly and just delete toValidLegalEntities

nicoprow (changes requested):
I can't execute the GET input legal-entities endpoints without error. See comments for improvement suggestions. 

Also the tests are not adapted. Currently, the tests for GET input legal-entities still configure the Wiremock for SaaS-Service Mocking. But that is no longer needed when fetching the business partners from the database directly.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/SiteService.kt
```diff
@@ -140,6 +140,8 @@ class SiteService(
         val sitesSaas = toSaasModels(sites)
         saasClient.upsertSites(sitesSaas)
 
+        sitePersistenceService.persistSitesBP(sites)
+
```
nicoprow, line 144: persist logic will be doubled like this

nicoprow (changes requested):
Change to the site service is unnecessary in my opinion. Also, it tries to persist the site twice now.

Issue #155, opened 2023-05-12
Pool: Remove SaaS import logic


nicoprow:
Currently, the Pool contains integration logic with a SaaS provider to import business partner data. However, the Pool should be independent of a specific SaaS provider. Therefore, we should remove that logic from the Pool.

For future providers who want to integrate against the Pool, they can use the Pool API to create and update business partners.

fabiodmota:
Done on https://github.com/eclipse-tractusx/bpdm/pull/206

PR #206, opened 2023-05-26
Feat/removing saas import logic

fabiodmota:
Removing Saas component and logic
Replacing unit testing with calls to create bpns


nicoprow (changes requested):
Looks good but it still has the [application-saas.properties](https://github.com/eclipse-tractusx/bpdm/blob/main/bpdm-pool/src/main/resources/application-saas.properties) in the resources. If this file is deleted, pull request is ready to merge

nicoprow:
closes #155

PR #223, opened 2023-06-01
Feat/Pool API: Legal Entity Response should include Legal Address 

rainer-exxcellent:


nicoprow:
Already merged with rebased branch: #229

Issue #209, opened 2023-05-30
README: Remove SaaS from install instructions


nicoprow:
Currently the INSTALL.md file contains descriptions about the saas profile. Since we removed SaaS from the pool we need to delete this from the documentation.

Issue #190, opened 2023-05-24
Run Instruction from Parent POM


nicoprow:
People who run an application for the first time need to do so from parent pom file and not from the child poms.

README should state this and show how to run an BPDM application from parent.

Issue #189, opened 2023-05-24
Update default Swagger URL


nicoprow:
The INSTALL.md contains an outdated swagger URL.

This URL should be replaced by the new default URL from the config.

PR #230, opened 2023-06-02
fix(gate/pool): Updated Install and read me instructions

cezaralexandremorais:
Changed information from Install.md and application properties, removing not needed SaaS instructions, Update Swagger-UI url, and wrote run instructions saying that the project must be run from root pom file

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -7,6 +7,7 @@ This file contains information on how to configure and run the BPDM applications
 BPDM Pool is a SpringBoot Kotlin software project managed by Maven.
 
 The project can be run with the following command: `mvn clean spring-boot:run`
+The command should be run from the root poom file
 
 ### Prerequisites
 
@@ -15,16 +16,15 @@ The project can be run with the following command: `mvn clean spring-boot:run`
 * PostgreSQL 14.2
 * OpenSearch 2.1.0
 * Keycloak 17.0.0 (with enabled `auth` profile)
-* Connection to an SaaS for the sharing process (with enabled `saas` profile)
 
 When running, the project requires a Postgresql database and an Opensearch instance to be available to connect to.
 Per default configuration the application expects postgres to run on `localhost` on port `5432`.
 Opensearch needs to run on `localhost` on port `9200` on default.
 
-You can find and edit the default configuration for the Pool in the `application.properties`,  `application-auth.properties` and `application-saas.properties`
+You can find and edit the default configuration for the Pool in the `application.properties`,  `application-auth.properties`
 files in the `resources` folder.
 
-The REST API documentation can be accessed at http://localhost:8080/api/swagger-ui.
+The REST API documentation can be accessed at http://localhost:8080/ui/swagger-ui/index.html.
 
 ### Profiles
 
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -7,6 +7,7 @@ This file contains information on how to configure and run the BPDM applications
 BPDM Pool is a SpringBoot Kotlin software project managed by Maven.
 
 The project can be run with the following command: `mvn clean spring-boot:run`
+The command should be run from the root poom file
```
nicoprow, line 10: The project can't be run with that command from the parent pom. You would need to specify the module as a target. We can maybe tell here that we can build this whole project from the parent pom and then depending on which application we want to run we need to specify the respective module

INSTALL.md
```diff
@@ -33,11 +33,11 @@ Here you can find core application configuration such as Swagger documentation,
 Furthermore, here you can find the configuration for the connection to the Spring datasource (currently, developed against PostgreSQL) and Opensearch.
 
 You can also run the project with Spring profiles to enable additional components on top of the default configuration.
-Currently, the BPDM Pool offers the profiles `auth` and `saas`.
+Currently, the BPDM Pool offers the profiles `auth`.
```
nicoprow, line 36: Should be singular expression here

INSTALL.md
```diff
@@ -33,11 +33,11 @@ Here you can find core application configuration such as Swagger documentation,
...
 In order to run the application with a specific profile you can use the appropriate maven flag `Dspring.profiles.active`.
 
 For example, the command `mvn clean spring-boot:run -Dspring.profiles.active=auth` starts the application with additional `auth` configuration enabled.
-You can also run several profiles at once, of course: `mvn clean spring-boot:run -Dspring.profiles.active=auth,saas`.
+You can also run several profiles at once, of course: `mvn clean spring-boot:run -Dspring.profiles.active=auth`.
```
nicoprow, line 40: Doesn't make any sense anymore since we only have one profile to run

bpdm-gate/src/main/resources/application.properties
```diff
@@ -28,17 +28,6 @@ bpdm.api.upsert-limit=100
...
-bpdm.saas.host=${BPDM_SAAS_HOST:http://localhost:1234}
-bpdm.saas.storage=${BPDM_SAAS_STORAGE:storage_id}
-bpdm.saas.datasource=${BPDM_SAAS_DATASOURCE:datasource_id}
-bpdm.saas.legal-entity-type=LEGAL_ENTITY
-bpdm.saas.site-type=ORGANIZATIONAL_UNIT
-bpdm.saas.address-type=BP_ADDRESS
-bpdm.saas.api-key=${BPDM_SAAS_KEY:}
-bpdm.saas.data-exchange-api-url=/data-exchange/rest/v4/storages/${bpdm.saas.storage}
-bpdm.saas.data-clinic-api-url=/data-clinic/rest/storages/${bpdm.saas.storage}
-bpdm.saas.sharing-timeout=2d
```
nicoprow, line 41: You are deleting the default values here for the properties. The actual config properties are still there as a data class in the code. Better keep those properties in for now and we make a separate issue/pull request to delete the necessary config properties from the code

INSTALL.md
```diff
@@ -7,6 +7,8 @@ This file contains information on how to configure and run the BPDM applications
 BPDM Pool is a SpringBoot Kotlin software project managed by Maven.
 
 The project can be run with the following command: `mvn clean spring-boot:run`
```
nicoprow, line 9: Instructin should be similar to:

root: `mvn clean install`

Go to the module subfolder you like to run and use the following command: `mvn spring-boot:run`

nicoprow:
Pull Request solves issues #209 #190 #189

Issue #241, opened 2023-06-05
Gate API: Merge Changelog Endpoints in Gate


nicoprow:
Currently the Gate has two changelog endpoints for querying and filtering the changelog. One for filtering by modifiedAfter time and list of external-ids. The other filters by modifiedAfter time and lsa-Type. Since external-ids are not unique to lsa-type it makes sense to merge these two endpoints together and apply all those filters together.

rainer-exxcellent:
Looks good.
In the ChangeLogSearchRequest could be remarked, that when externalIds or the lsaTypes are empty the parameters are ignored

PR #242, opened 2023-06-05
Gate API: Overhaul Changelog Endpoint

nicoprow:
Merge the two changelog endpoints of the BPDM Gate and create a new endpoint which contains the filters fromTime, external-ids and lsaTypes. Fixes issue #241


rainer-exxcellent:
Looks good.
In the ChangeLogSearchRequest could be remarked, that when externalIds or the lsaTypes are empty the parameters are ignored

Issue #236, opened 2023-06-05
Gate API: Remove BPN from Business Partner Input


nicoprow:
It makes no sense that a sharing member can set the BPN in the Gate as it is a derived value from our golden record process. It is unclear in what way a self-set BPN would help the overall process, if it all. Therefore, the BPN field should be removed in all Business Partner Input DTOs.

Likewise the BPN can be removed from the entities as well as it is part of the Sharing State at the moment.

nicoprow:
BPN should not be removed from the entities as it would be needed for the output DTOs

PR #244, opened 2023-06-06
fix(gate) Removed BPN Property from Gate

cezaralexandremorais:
Removed BPN property from LogisticAddress , LegalEntity and Site entities, input request and response, and updated unit tests to work without BPN.
Created a Migration file to drop BPN column from Site, LegalEntity and LogisticAddress


nicoprow:
fixes issue #236

nicoprow (approved):

PR #251, opened 2023-06-07
Add new Roles

fabiodmota:
Added roles to BusinessPartner classes
Introduced Role enumeration and integrated with BusinessPartner entities


nicoprow:
Duplicate of #257

PR #286, opened 2023-06-14
Fix - Legal/Main Address Changelogs on Update/Create

### LINKED ISSUE #270: Gate: No Legal/Main Address Changelogs
Currently, when the legal/main address is updated through an update on the legal entity/site, no changelog entry is created.

Expected:

 When the legal entity is updated there should be a changelog entry for the legal address. Likewise there should be a changelog entry for the main address if a site is updated.

alexsilva-CGI:
When the legal entity is updated there should be a changelog entry for the legal address. Likewise there should be a changelog entry for the main address if a site is updated.

Related Issue: fix #270


nicoprow (approved):
looks good to me

Issue #263, opened 2023-06-10
NameParts: Replace Array with Collection Type


nicoprow:
Array is a very specialized type in Kotlin and can better be replaced by the normal Collection type. Currently some namePart fields (legalNameParts) are of type Array and should be converted to the Collection Type. The resulting JSON will be unaffected.

fabiodmota:
Waiting for review on https://github.com/eclipse-tractusx/bpdm/pull/287

PR #287, opened 2023-06-14
fix(Gate-Bridge): Replace Array with Collection Type #263

fabiodmota:
Refactor: Replace Array with List in NameParts field


rainer-exxcellent:
Approved: for me it looks good

nicoprow (approved):

PR #293, opened 2023-06-15
Fix - Removal of unused columns in the persistence

### LINKED ISSUE #273: Gate: Delete Legal Entity Currentness Column
In the Gate currentness is not a concept. Therefore the respective column  for currentness in the legal entity should be deleted.

### LINKED ISSUE #274: Gate Persistence: Delete Identifier Entities
In the Gate we do not keep meta data information like identifier types. The identifier for legal entities or addresses is a free text field that will get evaluated and resolved by the golden record process in the next step. The Gate does not need to keep identifier types of any kind in its persistence layer. Therefore, the legal entity identifier type and address identifier type entities can be deleted.

### LINKED ISSUE #275: Gate Persistence: Delete "siteExternalId" Column
The logistic address entity in the Gate contains the column siteExternalId which is not needed, as the a relation with a site is expressed by the site_id column. siteExternalId can there therefore be deleted from the entity and database.

### LINKED ISSUE #276: Gate Persistence: Delete column "administrativeAreaLevel4"
In the Gate the physical postal address contains the column administrativeAreaLevel4 which is not longer needed. The columns should be deleted from the database and entity.

### LINKED ISSUE #277: Gate Persistence: Delete districtLevel2 Column
In the physical postal address the column districtLevel2 is not needed anymore and should be deleted

### LINKED ISSUE #278: Gate Persistence: Delete Region
The Gate does not keep meta data, therefore the Region entity can be deleted. Also any reference to a region should be replaced by string type that gets resolved in the Gold Record process in the next step.

alexsilva-CGI:
---
name: Removal of unused columns in the persistence
---

## Description

This PR removes unused and unnecessary columns in the Entities and Database from the Gate.

Associated issues: fix #273 / fix #274 / fix #275 / fix #276 / fix #277 / fix #278

## Pre-review checks


nicoprow (changes requested):
Looks good but one thing:

I saw that the migration file versions in the Gate are a bit off. The idea is that the first three numbers should match the app version and the last numberis the increment. In the latest migration file I set the migration file version to 4_0_0_0 indicating the current app version and the increment 0. Following this versioning the next migration file should be versioned 4_0_0_1

Issue #261, opened 2023-06-10
Gate: Remove Remaining SaaS Components


nicoprow:
Currently the Gate still contains some SaaS configuration and logic even though this logic is not used anymore in the endpoints. It is time to delete the remaining classes. In general everything related to "SaaS" should be removed from the Gate, including:

SaasConfig
SaasConfigProperties (and the values in the properties and test-properties files)
The SaaS exceptions
BusinessPartnerSaasWithBpn
BusinessPartnerSaasWithExternalId
InputSaasMappingService
OutputSaasMappingService
SaasClient
SaasDtoToSaasAddressMapping
SaasRequestMappingService
SaasDtoToSaasAddressMappingTest
SaasValues

PR #302, opened 2023-06-22
fix(gate): removed SaaS from gate

cezaralexandremorais:
Removed Remaining SaaS from Gate, and fix the issue #261


nicoprow (changes requested):
The content itself looks good to me but please consider the limit changed lines of code. 

Please split the pull request in such a way that the number of lines of code is around 1000 at most.

PR #305, opened 2023-06-26
Fix/remove saas classes from pool and common lib part 2

cezaralexandremorais:
removed some classes from common lib and pool that are no used any more, this commit is a second part of a bigger pull request that was split in 3
the other ones are #304  and #306


nicoprow (approved):
Looks good, needs to wait for #304 for merge

nicoprow (approved):

PR #306, opened 2023-06-26
Fix/remove saas from pool and common lib part 3

cezaralexandremorais:
removed some classes from common lib and pool that are no used any more, this commit is a third part of a bigger pull request that was split in 3
the other ones are #304  and #305


cezaralexandremorais:
need to wait for #305

cezaralexandremorais:
> Not sure what is the content of this pull request. Looks like a duplicate commit history and duplicate of #306
 this commit is a third part of a bigger pull request that was split in 3
the other ones are #304  and #305

nicoprow (approved):

PR #308, opened 2023-06-26
Fix/remove saas gate 2

cezaralexandremorais:
removed some saas classes and logic from gate


cezaralexandremorais:
need to wait for #307

nicoprow (changes requested):
Content Looks good but clean up the commit history please so you don't have two commits with the same description

nicoprow (approved):

PR #312, opened 2023-06-27
Fix/name parts persist

### LINKED ISSUE #271: Gate: Persist Name Parts
In the Gate, the DTOs to update legal entity, site and address partners contain the name parts field. However, the entities in our persistence layer just have plain String names. We should save the name parts from the DTOs in the database by replacing the name columns with name part collections in the entities.

alexsilva-CGI:
---
### Title: Persistence of Name Parts
---

## Description
This PR adds an table to be able to persist the Name Parts in Legal Entities, Addresses and Sites. On Sites and Logistic Addresses the name column was also removed. Respective mapping was done also.

Related issues: fix #271 

---
### Pre-review checks


nicoprow (changes requested):
Logic and structure for name parts looks good. But we lose the current names of the business partners. Please migrate the currently existing business partner names so that they appear as the first entry of the name parts array. 

Also, please delete the legalName from the LegalEntity as well. The Legal Entity should only have nameparts and the legalShortName.

nicoprow (changes requested):
I receive an error during database migration. The inserted name parts during migration do not have an ID (nor timestamps). 

Also I didn't see that the Legal Entity names are migrated to name parts

nicoprow (changes requested):
Migration works just needs to resolve the conflicts and we are good to go

nicoprow (approved):

PR #317, opened 2023-07-04
Gate Persistence: Legal Form should be optional

### LINKED ISSUE #272: Gate Persistence: Legal Form should be optional
The Gate Entity for the legal entity requires the legal form to be set (column is not nullable).

Expected
The legal form identifier should be optional.

SujitMBRDI:
---
title: 'Gate Persistence: Legal Form should be optional'
---

## Description
This pull request makes column 'legal_form_id' from table 'bpdmgate.legal_entities' to accept Null values. And making Legal Form as optional attribute for Legal Entity on Gate Service.

Fixes #272 

## Pre-review checks

### DIFF
--- a/bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/LegalEntity.kt
+++ b/bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/LegalEntity.kt
@@ -38,8 +38,8 @@ class LegalEntity(
     @Column(name = "name_shortname")
     var shortName: String?,
 
-    @Column(name = "legal_form_id", nullable = false)
-    var legalForm: String?,
+    @Column(name = "legal_form_id")
+    var legalForm: String? = null,
 
     @Column(name = "data_type")
     @Enumerated(EnumType.STRING)
--- a/bpdm-gate/src/main/resources/db/migration/V4_0_0_6__make_legal_form_id_optional.sql
+++ b/bpdm-gate/src/main/resources/db/migration/V4_0_0_6__make_legal_form_id_optional.sql
@@ -1,0 +1,3 @@
+-- Description: Make 'legal_form_id' column optional in the 'legal_entities' table
+-- Alter the 'legal_form_id' column to allow NULL values
+ALTER TABLE legal_entities ALTER COLUMN legal_form_id DROP NOT NULL;


bpdm-gate/src/main/resources/db/migration/V4_0_0_5__make_legal_form_id_optional.sql
```diff
@@ -0,0 +1,3 @@
+-- Description: Make 'legal_form_id' column optional in the 'legal_entities' table
+-- Alter the 'legal_form_id' column to allow NULL values
+ALTER TABLE bpdmgate.legal_entities ALTER COLUMN legal_form_id DROP NOT NULL;
```
nicoprow, line 3: Please remove the schema name here. Without the explicit schema name we would be more flexible in the configuration on where these migration files are applied.

nicoprow (changes requested):
Looks good so far, just needs a small change in the migration file.

Other than that the commit message format needs to be adapterd. Our commit message format is inspired by angular2 format: https://github.com/angular/angular/blob/main/CONTRIBUTING.md#commit

Scopes in our commit messages are not taken from a list though but can be freely chosen by you.

Regarding this, your commit message would be something like this: "fix(gate): ..."

bpdm-gate/src/main/resources/db/migration/V4_0_0_5__make_legal_form_id_optional.sql
SujitMBRDI: Removed schema name

SujitMBRDI:
> Looks good so far, just needs a small change in the migration file.
> 
> Other than that the commit message format needs to be adapterd. Our commit message format is inspired by angular2 format: https://github.com/angular/angular/blob/main/CONTRIBUTING.md#commit
> 
> Scopes in our commit messages are not taken from a list though but can be freely chosen by you.
> 
> Regarding this, your commit message would be something like this: "fix(gate): ..."

Thanks for providing syntax for adding commit message. I have adapted format in recent commit.
Also fixed schema name as  suggested.

nicoprow (changes requested):
Ok, I thought the failing test execution is a race condition problem but actually the migration versions are duplicated. On the current main there is now already a migration file with version 4_0_0_5, so you would need to increase it. Best is to rebase your branch to main first, then execute the tests again before force pushing

SujitMBRDI:
> Ok, I thought the failing test execution is a race condition problem but actually the migration versions are duplicated. On the current main there is now already a migration file with version 4_0_0_5, so you would need to increase it. Best is to rebase your branch to main first, then execute the tests again before force pushing

Done and ready to merge

PR #326, opened 2023-07-11
fix(gate): fixed the persistence of AddressIdentifier

cezaralexandremorais:
fixed the persistence of the identifiers collection on LogisticAddress


nicoprow (changes requested):
I tried the feature and when updating/creating addresses directly it works. But when updating/creating legal entities or sites, the identifiers are not saved in the legal addresses and main addresses.

nicoprow (changes requested):
Works as intended  but I just noticed in the Tests you ignore the identifiers fields when comparing. I think it's better to compare them as well and really cover the identifiers in the tests again.

cezaralexandremorais:
> Works as intended but I just noticed in the Tests you ignore the identifiers fields when comparing. I think it's better to compare them as well and really cover the identifiers in the tests again.

Done, made some changes to validate identifiers on tests

nicoprow (approved):

PR #335, opened 2023-07-17
feat(gate): created endpoint for Regions Metadata

### LINKED ISSUE #267: Pool API: Add Region Meta Data
The Pool API should have new meta data endpoints to add and query region information.

cezaralexandremorais:
Created endpoint to persist and get the Region metadata

Fixes #267


nicoprow (changes requested):
RegionDto and RegionRequest are the same. Best to delete RegionRequest and just use RegiondDto instead

cezaralexandremorais:
> RegionDto and RegionRequest are the same. Best to delete RegionRequest and just use RegiondDto instead

done

nicoprow (approved):

PR #341, opened 2023-07-20
feat: Bridge tests for recorded sync state

### LINKED ISSUE #212: Bridge Dummy: Add integration tests for polling mechanism
The bridge dummy should include integration tests for the polling mechanism: 

- On the next polling should continue from the last entry it has polled before (save state on success)
- When an error happens during polling, the next polling should start from the same entry again (save state on error)

martinfkaeser:
SyncStateIT tests if during a sync the bridge polls the Gate changelogs starting at the correct timestamp.
The SyncRecord and SyncRecordService is improved.

Fixes #212


martinfkaeser:
@nicoprow @rainer-exxcellent 
We should discuss the changes to the SyncRecordService

nicoprow (approved):

PR #385, opened 2023-08-10
Feat - API mermaid models 

alexsilva-CGI:
---
Name: API mermaid models 
---

## Description

In this PR, new models based on OpenAPI specifications we're added using mermaid notation. They are divided into Pool and Gate, and themselves divided into Addresses, Legal Entities, Sites and the remaining endpoints.

Preview Charts here:

**Gate**  / **#### Deprecated ####**
- Sites --> https://www.mermaidchart.com/raw/7ac5b22f-18fe-4d8b-8d97-b6f8403f3dd2?version=v0.1&theme=light&format=svg
- Legal Entities --> https://www.mermaidchart.com/raw/641225a3-5d91-4c28-ab06-fbb60be0389b?version=v0.1&theme=light&format=svg
- Addresses --> https://www.mermaidchart.com/raw/cd562b8e-89e3-420b-b360-ebc1ec98595d?version=v0.1&theme=light&format=svg
- Changelog and Sharing --> https://www.mermaidchart.com/raw/f670bdae-9eb8-4aaf-b606-675736d465ee?version=v0.1&theme=light&format=svg

**Pool** / **#### Deprecated ####**
- Sites --> https://www.mermaidchart.com/raw/3107826c-4821-4573-9738-f1733b4715fe?version=v0.1&theme=light&format=svg
- Remaining ones available in .MD format

## Pre-review checks

### Additional information

- To request a review, check out [who's involved](https://projects.eclipse.org/projects/automotive.tractusx/who) to see a list of contributors and committers
- Use the [mailing list](https://accounts.eclipse.org/mailing-list/tractusx-dev) if you are unsure, who to contact, or if want to engage in a general discussion
- Check out our guide on [how to contribute](https://eclipse-tractusx.github.io/docs/oss/how-to-contribute)
- Check out our [Release Guidelines](https://eclipse-tractusx.github.io/docs/release)
- Check out the [Eclipse Foundation contribution guidelines](https://www.eclipse.org/projects/handbook/#contributing)


Sebastian-Wurm (changes requested):
I guess, the diagrams would be easier consumable, if:
- we split them by OpenAPI controller ("site-controller" as sub folder), resource and request vs. response ("output-sites-response.md" as file) --> we would not need the notes then
- we would remove attributes where associations exist and indicate the name as a role of the association (like for geographicCoordinates)
- we would indicate cardinalities

We would end up with smaller diagrams and could visualize an OpenAPI resource request or response with one file.

Question: are these mermaid diagrams created by hand or generated from the OpenAPI spec in an automated way? What if the OpenAPI Spec changes?

alexsilva-CGI:
Hello @Sebastian-Wurm. Although they don't have cardinality yet, Did some changes on the models as you requested. I split them by controller Pool/Gate (Site, Address, Legal Entity, etc). 

- As for the generation of this diagrams, I used an library [openapi-mermaid](https://www.npmjs.com/package/openapi-mermaid), but it's pretty limited. I was able to generate the diagrams using the OpenAPI .json files present on the project, but what happens is that generates the full diagram for Gate/Pool, and to get them readable I needed to split them manually. 

- Regarding the OpenAPI Specs changing, the only solution is to update them manually I think.

Could you take a look and tell me if something needs some change? Thanks

alexsilva-CGI:
Hello once again @Sebastian-Wurm. Will answers your questions.

- So, it is an option to generate a diagram. For example regarding the Gate, we could split manually and then generate an site, address, gate and changelog diagram. After that, we need to split them manually like I did here I think, as it would be less confusing. It does require some work, but it is doable. Firstly what was talked about was to generate these diagrams for a base model, and then update them manually after every major change, for example.
- Didn't take that into account. I'm working with the current OpenAPI Json file available on the project. Probably isn't up to date as well.
- Indeed, didn't remember of that case. Will take a look into that as well.
- Yeah, it definitely can be a little confusing. Some response/request object will use this DTOs and code wise it works, but as you said, in the models it should have a name in accordance with the other objects to be more "cleaner". I think something can be done about that, maybe some refactoring on the code.

- Now, in regards to the cardinality. Is it necessary for this type of diagram? It will require some work, as it will need to be added manually, but it can be done :)

Thanks,
Alexandre

Sebastian-Wurm:
> Hello once again @Sebastian-Wurm. Will answers your questions.
> 
> * So, it is an option to generate a diagram. For example regarding the Gate, we could split manually and then generate an site, address, gate and changelog diagram. After that, we need to split them manually like I did here I think, as it would be less confusing. It does require some work, but it is doable. Firstly what was talked about was to generate these diagrams for a base model, and then update them manually after every major change, for example.
> * Didn't take that into account. I'm working with the current OpenAPI Json file available on the project. Probably isn't up to date as well.
> * Indeed, didn't remember of that case. Will take a look into that as well.
> * Yeah, it definitely can be a little confusing. Some response/request object will use this DTOs and code wise it works, but as you said, in the models it should have a name in accordance with the other objects to be more "cleaner". I think something can be done about that, maybe some refactoring on the code.
> * Now, in regards to the cardinality. Is it necessary for this type of diagram? It will require some work, as it will need to be added manually, but it can be done :)
> 
> Thanks, Alexandre

@alexsilva-CGI, @nicoprow:
- OpenAPI specification from latest Swagger UI contains the change I've mentioned above. Also, new "Dto" suffixes have been introduced there in comparison to the version I've used earlier.
- I could live without cardinalities for now. If the idea is to take these data models for the description of the API standards, we should include the cardinalities and remove attributes, where an association exists.

nicoprow:
Contributes to issue #475

PR #394, opened 2023-08-16
feat(docs): postman collection for automated api tests

SujitMBRDI:
## Description

This pull request contains postman collection for automated tests in BPDM. where user can select environment as per the need and perform actions like creating metadata, query metadata, creating business partner referencing metadata, query business partner, syncing new business partners to pool, query business partner in pool, updating business partner, syncing again, query business partner update in one click. 

## How to Test

Create local or any specific environment variables then import BPDM Automated Test.postman_collection.json into postman.
After importing, simply run collection by selecting particular intended environment. Result will show passed and failed test cases in postman.

### DIFF
--- a/docs/postman/automated-test/BPDM Automated Test.postman_collection.json
+++ b/docs/postman/automated-test/BPDM Automated Test.postman_collection.json
@@ -1,0 +1,1110 @@
+{
+	"info": {
+		"_postman_id": "b0c17786-b741-4a06-a4b2-a0800981e2f9",
+		"name": "BPDM Automated Test",
+		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json",
+		"_exporter_id": "26818013"
+	},
+	"item": [
+		{
+			"name": "Create Metadata LE Identifier Type",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"exec": [
+							"pm.test(\"Status code is 200 or 409\", function () {",
+							"    // Get the response status code",
+							"    var statusCode = pm.response.code;",
+							"",
+							"    // Check if the status code is either 200 or 409",
+							"    pm.expect(statusCode).to.be.oneOf([200, 409]);",
+							"});",
+							"pm.test(\"Check Legal Entity Identifier\", function () {",
+							"    var responseBody = pm.response.text();",
+							"    var customLEId = pm.collectionVariables.get(\"customLEId\");",
+							"    pm.expect(responseBody).to.include(customLEId);",
+							"});",
+							""
+						],
+						"type": "text/javascript"
+					}
+				},
+				{
+					"listen": "prerequest",
+					"script": {
+						"exec": [
+							"var uuid = require('uuid');",
+							"var customLEId = \"CUSTOM_LE_ID_\" + uuid.v4();",
+							"pm.collectionVariables.set('customLEId', customLEId);",
+							""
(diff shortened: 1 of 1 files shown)


SujitMBRDI:
> The type of requests and order makes a lot of sense. But it is very difficult to use the collection for regression tests as the user always needs to change the IDs of the DTOs in order to reuse the requests. To make this collection more usable I would suggest to extract technical keys, BPNs and external ids into variables. Technical keys and external ids can then be set to random strings before the respective request is executed. Created BPNs should be saved to its variable automatically after the request. This way we can execute the collection without adjusting all variables first.

I have added technical keys and external ids to variable which can dynamically produce them but for every run huge data will be created.

docs/postman/automated-test/BPDM Automated Test.postman_collection.json
```diff
@@ -0,0 +1,1112 @@
...
+							""
+						],
+						"type": "text/javascript"
+					}
+				},
+				{
+					"listen": "prerequest",
+					"script": {
+						"exec": [
+							"var customLEId = \"CUSTOM_LE_ID_\" + Math.random().toString(36).substring(2, 12);",
```
nicoprow, line 37: You can make your life easier by using the uuid functionality:

var uuid = require('uuid');
var customLEId =  uuid.v4();

docs/postman/automated-test/BPDM Automated Test.postman_collection.json
```diff
@@ -0,0 +1,1112 @@
...
+					}
+				},
+				{
+					"listen": "prerequest",
+					"script": {
+						"exec": [
+							"var customLeIdValue = pm.collectionVariables.get(\"customLeId\");",
+							"",
+							"// Construct the request URL with query parameters",
+							"var requestUrl = `${pm.variables.get(\"Pool-Host\")}/api/catena/identifier-types?businessPartnerType=LEGAL_ENTITY&technicalKey=${customLeIdValue}`;",
```
nicoprow, line 154: Why setting the request URL here? It is already defined in the collection

nicoprow (changes requested):
Pre-Request Scripts and Tests are now used and it improves the usability of the collection.

I would suggest to improve the way how random technical keys are generated. Also I would not set the request urls automatically before request execution.

The tests now also check response values which is a nice addition. Still, it should be clear to the user that his Postman collection is not meant to be run fully automatically and still needs a user to validate responses.

docs/postman/automated-test/BPDM Automated Test.postman_collection.json
SujitMBRDI: Updated with UUID, Thank you.

docs/postman/automated-test/BPDM Automated Test.postman_collection.json
SujitMBRDI: Removed :)

SujitMBRDI:
> Pre-Request Scripts and Tests are now used and it improves the usability of the collection.
> 
> I would suggest to improve the way how random technical keys are generated. Also I would not set the request urls automatically before request execution.
> 
> The tests now also check response values which is a nice addition. Still, it should be clear to the user that his Postman collection is not meant to be run fully automatically and still needs a user to validate responses.

As we discussed, I have performed changes and squashed commits too

nicoprow (approved):

Issue #407, opened 2023-08-22
Refactor: Reduce deeply-nested structures in LogisticAddress DTOs


martinfkaeser:
We want to replace nested LogisticAddress DTOs using JsonUnwrapped with interfaces for commonly used fields and properties.

Example:

Convert the nested DTOs to interfaces and let PhysicalPostalAddressVerboseDto implement these interfaces. Remove nested DTOs. Adjust services, test objects and tests as needed.

PR #408, opened 2023-08-23
Refactor: Nested logistic address DTOs

martinfkaeser:
Reduce deeply-nested structures using JsonUnwrapped in LogisticAddress DTOs with interfaces for commonly used fields.
Introduced interfaces IBasePostalAddressDto and IBasePhysicalPostalAddressDto as replacement.

Solves #407

Based on https://github.com/eclipse-tractusx/bpdm/pull/398, so this must be merged first


martinfkaeser:
I made an alternative version where I added an extra interface for each class in the hierarchy. 
This makes it extra clear what fields are added to each subclass. The OpenAPI field annotations show only up in the interfaces. The actual classes only define the constructor with the exact order and maybe default parameters.
I'm curious what solution you prefer, I can still drop the second commit.

bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/IBasePostalAddressDto.kt
```diff
@@ -19,22 +19,20 @@
...
 
-import com.neovisionaries.i18n.CountryCode
 import io.swagger.v3.oas.annotations.media.Schema
 import org.eclipse.tractusx.bpdm.common.dto.openapidescription.PostalAddressDescription
 
-data class BasePostalAddressDto(
+/**
+ * Contains the common fields for all PostalAddress-DTOs, whether Physical or Alternative respectively Pool or Gate.
+ */
+interface IBasePostalAddressDto {
```
nicoprow, line 28: Should we really include this interface? It makes sense from a technical perspective as it contains common fields across all addresses but the fields feel just so... random. As a concept for a base postal address the interface really is not very expressive. I feel those three fields could also be added as duplicates in subclasses/subinterfaces as such an interface could feel more confusing. I'm open for discussion though

nicoprow (commented):
Using interfaces here is much better than the nested component structure we had before. But we should think about what kind of interfaces we want to introduce here

bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/IBasePostalAddressDto.kt
rainer-exxcellent: I thought exactly the same. Its only a technical Interface to group some fields without any unique context.
To remove it would also decrease the dept of the address hierarchy

rainer-exxcellent: When we remove the interface IBasePostalAddressDto perhaps it would make sense to create an IAlternativePostalAddressDto interface.

rainer-exxcellent:
For me it makes no sense to create a description object for descriptions that are only used once in an annotation like PostalAddressDescription.administrativeAreaLevel2

interface IBasePhysicalPostalAddressDto : IBasePostalAddressDto {

    @get:Schema(description = PostalAddressDescription.administrativeAreaLevel2)
    val administrativeAreaLevel2: String?

bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/IBasePostalAddressDto.kt
martinfkaeser: That's not a bad idea, IAlternativePostalAddressDto seems more helpful than IBasePostalAddressDto!

It still seems a little random because it's missing e.g. country that is essential to any address. But this fields unfortunately can't be included because it has a different type in the verbose and normal versions.

martinfkaeser: As decided together we now have only IBasePhysicalPostalAddressDto and IBaseAlternativePostalAddressDto as interfaces.

nicoprow (approved):

Issue #413, opened 2023-08-25
Pool: Duplicate Address Identifier Validation on New Business Partners


fabiodmota:
Expected behaviour:

When POSTing new legal addresses which have the same identifiers (value and type are the same) to the Pool, the Pool should reject these business partners due to duplicate identifiers.

 

Actual Behaviour:

The business partners which share the same identifiers are created instead without any validation error.

 

How to reproduce:

POST to the Pool create address entities endpoint two entities in one request which have the same identifiers and are otherwise valid entities. The address entities are then created successfully.

PR #414, opened 2023-08-25
Fix/duplicate identifier validation for address

fabiodmota:
## Description
Added service validation when a request is made with identifier with the same values at the address level.
Added unit test for testing the error return

https://github.com/eclipse-tractusx/bpdm/issues/413

## Pre-review checks

### DIFF
--- a/bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/response/PoolErrorCodes.kt
+++ b/bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/response/PoolErrorCodes.kt
@@ -33,7 +33,8 @@ enum class LegalEntityCreateError : ErrorCode {
     LegalFormNotFound,
     LegalEntityIdentifierNotFound,
     LegalAddressRegionNotFound,
-    LegalAddressIdentifierNotFound
+    LegalAddressIdentifierNotFound,
+    LegalAddressDuplicateIdentifier,
 }
 
 @Schema(description = "LegalEntityUpdateError")
@@ -43,21 +44,24 @@ enum class LegalEntityUpdateError : ErrorCode {
     LegalFormNotFound,
     LegalEntityIdentifierNotFound,
     LegalAddressRegionNotFound,
-    LegalAddressIdentifierNotFound
+    LegalAddressIdentifierNotFound,
+    LegalAddressDuplicateIdentifier
 }
 
 @Schema(description = "SiteCreateError")
 enum class SiteCreateError : ErrorCode {
     LegalEntityNotFound,
     MainAddressIdentifierNotFound,
-    MainAddressRegionNotFound
+    MainAddressRegionNotFound,
+    MainAddressDuplicateIdentifier
 }
 
 @Schema(description = "SiteUpdateError")
 enum class SiteUpdateError : ErrorCode {
     SiteNotFound,
     MainAddressIdentifierNotFound,
-    MainAddressRegionNotFound
+    MainAddressRegionNotFound,
+    MainAddressDuplicateIdentifier
 }
 
 @Schema(description = "AddressCreateError")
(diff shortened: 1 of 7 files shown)


martinfkaeser:
I found a case that is not handled yet.

The expected behaviour from the issue is:

> When POSTing new legal addresses which have the same identifiers (value and type are the same) to the Pool, the Pool should reject these business partners due to duplicate identifiers.

But now this fix only prohibits identifiers that are absolutely identical (regarding all 3 fields: type, value, issuingBody). 
If there are two requests where identifier type and value are equal, but issuingBody is different, this is not detected and both entities are accepted.

I have modified your LE test which should now detect this case too in branch `fix/Duplicate_Identifier_Validation_MKA` which you can cherry-pick from.

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/response/PoolErrorCodes.kt
```diff
@@ -66,7 +66,8 @@ enum class AddressCreateError : ErrorCode {
     SiteNotFound,
     LegalEntityNotFound,
     RegionNotFound,
-    IdentifierNotFound
+    IdentifierNotFound,
+    MainAddressDuplicateIdentifier,
```
martinfkaeser, line 70: Shouldn't it be just called `AddressDuplicateIdentifier`?

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/RequestValidationService.kt
```diff
@@ -225,6 +257,29 @@ class RequestValidationService(
         }.groupBy({ it.first }, { it.second })
     }
 
+
+    fun findDuplicateLegalEntityIdentifiers(
+        legalEntityRequests: List<LegalEntityDto>
+    ): Set<LegalEntityIdentifierDto> {
+        val allIdentifiers = legalEntityRequests.flatMap { it.identifiers }
+
+        return allIdentifiers.groupBy { it }
```
martinfkaeser, line 266: This should only consider the fields `type` and `value`. Maybe we could use data class `IdentifierCandidateKey` for that?

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/RequestValidationService.kt
```diff
@@ -78,13 +76,43 @@ class RequestValidationService(
...
+                        validateLegalEntityDuplicates(legalEntity,duplicateIdentifiers,request.index)
             validationErrors.map { Pair(request, it) }
 
         }.groupBy({ it.first }, { it.second })
     }
 
+    fun validateLegalEntityDuplicates(legalEntity: LegalEntityDto, duplicateIdentifiers: Set<LegalEntityIdentifierDto>, entityKey: String?) : Collection<ErrorInfo<LegalEntityCreateError>> {
+        val errorList = mutableListOf<ErrorInfo<LegalEntityCreateError>>()
+        duplicateIdentifiers.forEach { duplicate ->
+            if (legalEntity.identifiers.contains(duplicate)) {
```
martinfkaeser, line 89: Same as `findDuplicateLegalEntityIdentifiers` this should only consider the fields type and value.

martinfkaeser:
I think `validateLegalEntityUpdates()` and `validateAddressUpdates()` should also check for duplicate identifiers in the request, right?

martinfkaeser (changes requested):

martinfkaeser:
I think the address identifiers nested as legal address or site main address need to be checked as well.

fabiodmota:
i added a new comment to all new developments @martinfkaeser thank you for the feedback

martinfkaeser:
I combined the duplicate checks for identifiers in the DB and in the request.
Then I added a check & test for the case of a duplicate legal address identifier that is already in the DB.
Also added findDuplicates() extension function to CollectionUtils.

fabiodmota:
@martinfkaeser looks good to me , for me its resolved also

nicoprow (approved):

nicoprow (changes requested):
Just saw in the commit history, the second commit has a missing main commit description. Please add a header description to this commit to put the the additional description in context.

fabiodmota:
> Just saw in the commit history, the second commit has a missing main commit description. Please add a header description to this commit to put the the additional description in context.

Fixed , and refactored commit header @nicoprow

nicoprow (approved):

PR #436, opened 2023-09-04
doc(adr-5): adr for edc usage scenarios

gerlitmcoding:
this adr describes recommendations of EDC usage scenarios within bpdm solution together with value-added-service


Sebastian-Wurm:
Fine for me.

nicoprow (approved):

PR #448, opened 2023-09-11
Refactor: Remove common values in Pool test data

### LINKED ISSUE #484: Refactor: Replace Common-Values References in Verbose and Non-Verbose Test Objects
This issue aims to refactor the codebase by replacing common-values references in both verbose and non-verbose test objects for the "POOL" and "GATE" modules. This refactoring effort will improve code maintainability and set the foundation for managing test templates efficiently in the future.

## Scope
- Identify common-values references in test objects.
- Replace these references with more maintainable and reusable code.
- Ensure that both verbose and non-verbose test objects are updated as needed.

## Tasks
- [ ] Create/Refactor Values-Object for Pool 
- [ ] Create/Refactor Values-Object for Gate

SujitMBRDI:
## Description

In this pull request, removed commonsValue object which referred by test objects for creation of  business partner. Updated response and request values-object into BusinessPartnerVerboseValues and BusinessPartnerNonVerboseValues respectively  which can be utilised while running test classes.

Fixes #484


bpdm-pool/src/test/kotlin/org/eclipse/tractusx/bpdm/pool/util/ResponseValues.kt
```diff
@@ -36,111 +40,111 @@ import java.time.ZoneOffset
...
+        LegalEntityIdentifierVerboseDto("Another ID Value", RequestValues.identifierType2, "Body Y")
     private val identifier3 =
-        LegalEntityIdentifierVerboseDto(CommonValues.identifierValue3, RequestValues.identifierType3, CommonValues.issuingBody3)
+        LegalEntityIdentifierVerboseDto("An ID Value", RequestValues.identifierType3, "Official Z")
 
     val legalForm1 = LegalFormDto(
-        technicalKey = CommonValues.legalFormTechnicalKey1,
-        name = CommonValues.legalFormName1,
-        abbreviation = CommonValues.legalFormAbbreviation1,
+        technicalKey = "LF1",
```
rainer-exxcellent, line 60:
```suggestion
        technicalKey = RequestValues.legalForm1.technicalKey,
```

bpdm-pool/src/test/kotlin/org/eclipse/tractusx/bpdm/pool/util/ResponseValues.kt
```diff
@@ -36,111 +40,111 @@ import java.time.ZoneOffset
...
+        name = "Limited Liability Company",
```
rainer-exxcellent, line 61:
```suggestion
        name = RequestValues.legalForm1.name,
```

bpdm-pool/src/test/kotlin/org/eclipse/tractusx/bpdm/pool/util/ResponseValues.kt
```diff
@@ -36,111 +40,111 @@ import java.time.ZoneOffset
...
+        abbreviation = "LLC",
```
rainer-exxcellent, line 62:
```suggestion
        abbreviation =  RequestValues.legalForm1.abbreviation,
```

rainer-exxcellent (commented):
I think it would be better to derive the response values from the request values,
like in the suggestions i made

SujitMBRDI:
> I think it would be better to derive the response values from the request values, like in the suggestions i made

@rainer-exxcellent thanks for feedback, 
I agree and will update pr accordingly, @nicoprow, also suggested same but with otherway i.e. all request values should be derived from response values and it will be beneficial too.

SujitMBRDI:
> > I think it would be better to derive the response values from the request values, like in the suggestions i made
> 
> @rainer-exxcellent thanks for feedback, I agree and will update pr accordingly, @nicoprow, also suggested same but with otherway i.e. all request values should be derived from response values and it will be beneficial too.

@nicoprow, @rainer-exxcellent @martinfkaeser, I have updated changes we discussed and resolved conflict too.
Kindly again review the pull request.

nicoprow (changes requested):
Commits can be squashed, especially since the first commit is basically reverted through the subsequent commits. I don't see that the provided commits in the history make sense here. Either squash to one commit or divide the commits into something like name change, common value replacement in verbose values and common value replacement in non-verbose values.

I like nonVerboseValues better than requestValues as the name more accurately depicts how the values are actually used. But I'm still not on board with the name nonVerbose. Would like to have a more concise name differentating between verbose and non-verbose DTOs in the Pool. I suggest we have a discussion about it and come to a conclusion

SujitMBRDI:
> Commits can be squashed, especially since the first commit is basically reverted through the subsequent commits. I don't see that the provided commits in the history make sense here. Either squash to one commit or divide the commits into something like name change, common value replacement in verbose values and common value replacement in non-verbose values.
> 
> I like nonVerboseValues better than requestValues as the name more accurately depicts how the values are actually used. But I'm still not on board with the name nonVerbose. Would like to have a more concise name differentating between verbose and non-verbose DTOs in the Pool. I suggest we have a discussion about it and come to a conclusion

@nicoprow As we discussed, i kept  nonVerboseValues name for object class as it is and squashed commits to the one.
Kindly review pull request again

nicoprow (approved):

Issue #456, opened 2023-09-13
Gate: Implement service logic for generic business partner


martinfkaeser:
For the generic business partner model (GBPM, see #382) implement the service logic that is needed by the endpoint. 
This includes logic for upserting BPs and querying the input and output stages of BPs.

fabiodmota:
Should we close this one ? Its already merged @nicoprow

PR #457, opened 2023-09-13
Feat: Implement generic business partner service logic in Gate

martinfkaeser:
This implements the logic for persisting generic business partners and querying for existing BPs (stage input and output). This involves mapping between DTO and entity model. Some small changes to the DTO model were needed.

Solves #456

### DIFF
--- a/bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/AlternativePostalAddressDto.kt
+++ b/bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/AlternativePostalAddressDto.kt
@@ -27,7 +27,10 @@ import org.eclipse.tractusx.bpdm.common.model.DeliveryServiceType
 import org.eclipse.tractusx.bpdm.common.service.DataClassUnwrappedJsonDeserializer
 
 @JsonDeserialize(using = DataClassUnwrappedJsonDeserializer::class)
-@Schema(description = PostalAddressDescription.headerAlternative)
+@Schema(
+    description = PostalAddressDescription.headerAlternative,
+    requiredProperties = ["country", "city", "deliveryServiceType", "deliveryServiceNumber"]
+)
 data class AlternativePostalAddressDto(
 
     override val geographicCoordinates: GeoCoordinateDto?,
--- a/bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerMappings.kt
+++ b/bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerMappings.kt
@@ -1,0 +1,243 @@
+/*******************************************************************************
+ * Copyright (c) 2021,2023 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License, Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ******************************************************************************/
+
+package org.eclipse.tractusx.bpdm.gate.service
+
+import org.eclipse.tractusx.bpdm.common.dto.AlternativePostalAddressDto
+import org.eclipse.tractusx.bpdm.common.dto.ClassificationDto
+import org.eclipse.tractusx.bpdm.common.dto.GeoCoordinateDto
+import org.eclipse.tractusx.bpdm.common.model.StageType
+import org.eclipse.tractusx.bpdm.common.util.replace
+import org.eclipse.tractusx.bpdm.gate.api.model.*
+import org.eclipse.tractusx.bpdm.gate.api.model.request.BusinessPartnerInputRequest
+import org.eclipse.tractusx.bpdm.gate.api.model.response.BusinessPartnerInputDto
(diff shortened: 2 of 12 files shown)


bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -0,0 +1,109 @@
...
+    fun upsertBusinessPartnersInput(dtos: Collection<BusinessPartnerInputRequest>): Collection<BusinessPartnerInputDto> {
+        val stage = StageType.Input
+        val existingEntitiesByExternalId = getExistingEntityByExternalId(dtos, stage)
+
+        val savedEntities = dtos.map { dto ->
+            existingEntitiesByExternalId[dto.externalId]
+                ?.let { existingEntity ->
+                    businessPartnerMappings.updateBusinessPartner(existingEntity, dto)
+                    businessPartnerRepository.save(existingEntity)
+                        .also {
```
rainer-exxcellent, line 56: is the also here necessary? The it object is not used

martinfkaeser: I use it here because I want to return the saved entity and write the changelog only after saving.

rainer-exxcellent: I would reduce the complexity of the function by extracting two function. One for the update path and one for the insert path.
I think it would make it better readable

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerMappings.kt
```diff
@@ -0,0 +1,243 @@
...
+            isOwner = dto.isOwner,
+            bpnL = null,
+            bpnS = null,
+            bpnA = null,
+            postalAddress = toPostalAddress(dto.postalAddress)
+        )
+    }
+
+    fun updateBusinessPartner(entity: BusinessPartner, dto: BusinessPartnerInputRequest) {
+        entity.nameParts.replace(dto.nameParts.toMutableList())
```
rainer-exxcellent, line 98: I think in this case the toMutableList() is not neccessary

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -0,0 +1,112 @@
...
+            externalIds != null -> businessPartnerRepository.findByStageAndExternalIdIn(stage, externalIds, pageRequest)
+            else -> businessPartnerRepository.findByStage(stage, pageRequest)
+        }
+        return page.toPageDto(businessPartnerMappings::toBusinessPartnerInputDto)
+    }
+
+    fun getBusinessPartnersOutput(pageRequest: PageRequest, externalIds: Collection<String>?): PageDto<BusinessPartnerOutputDto> {
+        val stage = StageType.Output
+        val page = when {
+            externalIds != null -> businessPartnerRepository.findByStageAndExternalIdIn(stage, externalIds, pageRequest)
```
nicoprow, line 88: Behaviour in the Gate seems to be inconsistent regarding missing externalIDs. I would recommend to take the same behaviour as in the Pool: When the list of IDs is null or *empty* then return all objects.

nicoprow (changes requested):
Some remarks/points for discussion from my side

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -0,0 +1,112 @@
...
+@Service
+class BusinessPartnerService(
+    private val businessPartnerRepository: BusinessPartnerRepository,
+    private val businessPartnerMappings: BusinessPartnerMappings,
+    private val sharingStateService: SharingStateService,
+    private val changelogRepository: ChangelogRepository
+) {
+
+    fun upsertBusinessPartnersInput(dtos: Collection<BusinessPartnerInputRequest>): Collection<BusinessPartnerInputDto> {
+        val existingEntitiesByExternalId = getExistingEntityByExternalId(dtos, StageType.Input)
```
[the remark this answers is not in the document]
martinfkaeser: You're right

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
martinfkaeser: From a logical point of view handling null and empty collections differently makes a little more sense to me (null means no filtering; empty means I want only these specific business partners but there are none).
But I agree, from a practical perspective, the empty array case doesn't make a lot of sense and it makes the API probably easier to use, if null and empty are handled the same.
So I changed the check to isNullOrEmpty().

bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/AlternativePostalAddressDto.kt
```diff
@@ -27,7 +27,10 @@ import org.eclipse.tractusx.bpdm.common.model.DeliveryServiceType
 import org.eclipse.tractusx.bpdm.common.service.DataClassUnwrappedJsonDeserializer
 
 @JsonDeserialize(using = DataClassUnwrappedJsonDeserializer::class)
-@Schema(description = PostalAddressDescription.headerAlternative)
+@Schema(
+    description = PostalAddressDescription.headerAlternative,
+    requiredProperties = ["country", "city", "deliveryServiceType", "deliveryServiceNumber"]
```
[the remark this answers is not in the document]
martinfkaeser: I don't know, they are definitely needed to show the required-asterisks in the Swagger-UI.
In the Pool I see no fields marked as required at first glance.
But I agree, I really don't like this redundant requiredProperties annotation.
It would be really cool to get it derived from the field nullability.
I suggest we have a deeper look into it in another ticket.

nicoprow (approved):

Issue #428, opened 2023-09-01
Gate: Persist Generic Business Partner Input


nicoprow:
When a sharing member shares a business partner in a generic data model, this business partner needs to be persisted before it can be further processed.

Acceptance Criteria
* Database migration script for generic business partner
* Entity model for generic business partner
* Mapping from generic DTO model to entity model and back
* Upsert Service logic to persist generic business partner 

Dependencies
#384

PR #460, opened 2023-09-13
Fix: Generic business partner entity fields in Gate

martinfkaeser:
Changes to entity model (including DB migration):
- Make BusinessPartner.isOwner mandatory
- Rename DB fields for BusinessPartner.bpnL/bpnS/bpnA
- Make PostalAddress.addressType optional

Fixes for #428


nicoprow (changes requested):
Works but I'm not happy with the commit description as its very generic.

I would recommend adding at least the points in the pull request description to the commit description.

nicoprow (approved):

Issue #422, opened 2023-09-01
Orchestrator: Setup new Module


nicoprow:
With the new orchestration architecture we need to have a new application in the BPDM project that manages the processing of cleaning requests for business partner data.

### Tasks
 - [x] New pom module and Spring Boot application in the BPDM project called Orchestrator
 - [x] Module has a basic structure derived from the bpdm-common module
 - [x] Deployment files for the BPDM Chart
 - [x] Workflow to publish the CleaningServiceDummy Dockerfile
 - [x] Include the CleaningServiceDummy in Trivy

Dependencies
#377 
#424

gerlitmcoding:
Also add to deployment charts

PR #463, opened 2023-09-15
Feat: Module Orchestrator

rainer-exxcellent:
## Description

Create maven module for the orchestrator
At the moment no persistence to a database is configured.

### Solves
#422


nicoprow (approved):
Contributes to #422 by creating a new maven module, docker file and notice file

Issue #453, opened 2023-09-12
CleaningServiceDummy: Service and Setup Helmcharts


fabiodmota:
## Create Helm Charts Structure for CleaningService Module

We need to create a Helm chart structure for the new module, **CleaningService**. 
This Helm chart will define the deployment configurations, services, and other Kubernetes resources required for the **CleaningService** module.

### Acceptance Criteria:

- [x] A new Helm chart for **CleaningService** is created.
- [x] The chart should be compatible with our current Kubernetes setup.

### Tasks:

- [x] Create the base structure for the **CleaningService** Helm chart.
- [x] Define necessary Kubernetes resources (e.g., Deployment, Service, ConfigMap, etc.).
- [x] Test the Helm chart in a dev environment.

PR #468, opened 2023-09-15
Feat/add cleaning service helm chart

fabiodmota:
## Description

We added a new helm chart to the existing structure to be able to deploy the new cleaning service module.

issue: https://github.com/eclipse-tractusx/bpdm/issues/453

## Pre-review checks


nicoprow (changes requested):
There are some refactorings to the BPDM application in the commit. Please remove them and add them in a separate pull request as we would like to keep App and Chart contributions separate.

Introducing a new subchart is a change in the original chart. Therefore, you would need to increase the BPDM Chart alpha version for this. (Linting fails for this reason)

fabiodmota:
The Pull Request was changed based solely on the helm structure for the new cleaning service. 
Something detected that caused the workflow to fail was the fact that so many applications, when doing the helm lint test, failed with deployment problems due to resources, without the error being specified. 
The resource allocation for simple apps that don't yet have logic was reduced, and the problem was solved

Issue #472, opened 2023-09-18
Documentation: Pool Internal Runtime View


fabiodmota:
Create mermaid sequence diagrams showing the different services/components of the BPDM Pool and how they interact each other on these use cases:

- Create legal entity
- Update legal entity
- Search legal entity

Embed these diagrams into Arc42 document. Align with [Tim Gerlicher](https://jira.catena-x.net/secure/ViewProfile.jspa?name=gerlitm%40schaeffler.com)

PR #473, opened 2023-09-18
Docs: Add diagrams for pool components

fabiodmota:
## Description
Create mermaid sequence diagrams showing the different services/components of the BPDM Pool and how they interact each other on these use cases:

- Create legal entity
- Update legal entity
- Search legal entity

https://github.com/eclipse-tractusx/bpdm/issues/472

## Pre-review checks

### DIFF
--- a/docs/arc42/diagrams/create_legal_entity_diagram.md
+++ b/docs/arc42/diagrams/create_legal_entity_diagram.md
@@ -1,0 +1,67 @@
+# Sequence Diagram for Business Partner Data Management (BPDM) Pool
+
+```mermaid
+sequenceDiagram
+    autonumber
+    participant Client as Client
+    participant Controller as Controller
+    participant BPBuildService as BusinessPartnerBuildService
+    participant ValidationService as RequestValidationService
+    participant ChangelogService as ChangelogService
+    participant Repository as LegalEntityRepository
+
+    Note over Client,Controller: Path: /api/catena/legal-entities
+    Note over Client,Controller: Method: POST
+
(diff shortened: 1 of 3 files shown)


nicoprow (approved):
I think these are great diagrams, very detailed and explain a lot about the inner workings of the Pool. I have a small concern with this: The method names in the sequence diagrams could be replaced with a more abstract description. Since there are explanations in the below this is just a minor issue. Also, concepts change less often than method names and it would be a nightmare to keep this documentation up-to-date with the current names and method invocations in detail.

In general I think it's an excellent contribution, not sure though if we should have the aspiration to keep this documention the code (with names and so many details) up-to-date.

I would like to see @gerlitmcoding input on this. From my end, I would approve.

gerlitmcoding:
> I think these are great diagrams, very detailed and explain a lot about the inner workings of the Pool. I have a small concern with this: The method names in the sequence diagrams could be replaced with a more abstract description. Since there are explanations in the below this is just a minor issue. Also, concepts change less often than method names and it would be a nightmare to keep this documentation up-to-date with the current names and method invocations in detail.
> 
> In general I think it's an excellent contribution, not sure though if we should have the aspiration to keep this documention the code (with names and so many details) up-to-date.
> 
> I would like to see @gerlitmcoding input on this. From my end, I would approve.

As discussed one by one with @nicoprow it is good as you did it @fabiodmota. We will not update every change that comes with every pull request. We will check for each new "release/arc42" if the information are still up-to-date.

nicoprow (approved):

Issue #441, opened 2023-09-07
Gate: Map from LSA DTOs to Generic Business Partner DTO


rainer-exxcellent:
When a request comes in to upsert an LSA Business Partner, then it needs to be mapped to the generic model
and passed to the generic service.

All fields in the Dtos of the LegalEntity, Site and Address in the Gate has to be mapped to BusinessPartnerInputDto.

Issue #440, opened 2023-09-07
Gate: Replace LSA Typed Service Logic with Generic Service Logic


rainer-exxcellent:
When a request comes in to upsert an LSA Business Partner, then it needs to be mapped to the generic model
and passed to the generic service.
The generic service already persists the data. No LSA data is persisted anymore.
Upon receiveing a GET request for an LSA Business Partner the first step is to check if the data is of the requested type.
Then take the generic data and convert it to the LSA type.

nicoprow:
At this moment we do not intend to support L/S/A endpoints entering the golden record process through the orchestrator. For this reason attaching L/S/A to the service logic is not needed.

PR #477, opened 2023-09-19
Feat - Replacement of LSA Typed Service Logic with Generic Service Logic

alexsilva-CGI:
---
Name: Replacement of LSA Typed Service Logic with Generic Service Logic
---

## Description

In this PR, the its implemented the replacement of LSA Typed Service Logic with Generic Service Logic. Needed mapping is also created.

Solves #440 and #441

## Pre-review checks

### Additional information

- To request a review, check out [who's involved](https://projects.eclipse.org/projects/automotive.tractusx/who) to see a list of contributors and committers
- Use the [mailing list](https://accounts.eclipse.org/mailing-list/tractusx-dev) if you are unsure, who to contact, or if want to engage in a general discussion
- Check out our guide on [how to contribute](https://eclipse-tractusx.github.io/docs/oss/how-to-contribute)
- Check out our [Release Guidelines](https://eclipse-tractusx.github.io/docs/release)
- Check out the [Eclipse Foundation contribution guidelines](https://www.eclipse.org/projects/handbook/#contributing)


alexsilva-CGI:
Currently the only endpoints that we're not replaced with the generic logic we're the Upsert Output on the LSA types, as there is no logic for that present on the generic. This means that an upsert of output data is not possible as there is no input data present on the non generic model database.

nicoprow (changes requested):
The Persistence-Services are obsolete now and can be deleted.

There are some questions about the behaviour of the L/S/A endpoints in relation to the generic endpoints. I think our behaviour should be like that:
A business partner type can change (over L/S/A or generic)
Over L/S/A endpoints you can query the business partners which are valid L/S/A. Valid means that they have valid existing parents. If a bsuiness partner has an address type for L/S/A but missing valid parents this, an error is returned instead.

I believe that is the most flexible solution.

nicoprow:
At this moment we do not intend to support L/S/A endpoints entering the golden record process through the orchestrator. For this reason attaching L/S/A to the service logic is not needed.

Issue #476, opened 2023-09-19
Gate: Only External-ID in Generic Business Partner Input is required 


nicoprow:
When upserting generic business partner data input only the externalId is required. All other fields are optional. If the externalId is empty or null it should return an error.

### Tasks

- [x] Make city optional
- [x] Make postal address optional
- [x] Make country optional
- [x] Other fields in subobjects like Identifiers, Classifications and so on

martinfkaeser:
So, as I understand a `BusinessPartnerInputRequest` doesn't need a `postalAddress`. But if there is one, `physicalPostalAddress` is still required inside of it, right?
And `PhysicalPostalAddressGateDto` doesn't have any required fields?
What about `AlternativePostalAddressDto`? Going by the original ticket I guess it shouldn't have any required fields either, not even `deliveryServiceType` and `deliveryServiceNumber`?

And what about the output version? 
Ideally we wouldn't want to change the "required"-ness of fields, but that would mean introducing input and output versions of `PhysicalPostalAddressGateDto` and `AlternativePostalAddressDto`. 
In my opinion this is not worth the effort!

nicoprow:
> So, as I understand a `BusinessPartnerInputRequest` doesn't need a `postalAddress`. But if there is one, `physicalPostalAddress` is still required inside of it, right? And `PhysicalPostalAddressGateDto` doesn't have any required fields? What about `AlternativePostalAddressDto`? Going by the original ticket I guess it shouldn't have any required fields either, not even `deliveryServiceType` and `deliveryServiceNumber`?
> 
> And what about the output version? Ideally we wouldn't want to change the "required"-ness of fields, but that would mean introducing input and output versions of `PhysicalPostalAddressGateDto` and `AlternativePostalAddressDto`. In my opinion this is not worth the effort!
> 
> And `AlternativePostalAddressDto`, which now is used in the Pool and Gate needs to be split.

1. True, we don't have an overview of which fields are required in case someone actually gives a postal address, physical postal address or alternative postal address. I don't like the idea of having completely empty objects there. We should get into this and see if we can agree on some minimum required fields for these address objects.
2. Validating the integrity of the data is of course a given. We can of course have the same DTO for both, having nullable fields there and just use service logic to validate this. It is important that these fields are validated to not be null where applicable and at the same time the API user needs to understand that these fields are required. In my opinion, whatever we do these are the requirements.
[...]

Issue #425, opened 2023-09-01
Orchestrator API: Request Business Partner Cleaning Endpoint


nicoprow:
A BPDM Gate needs to request cleaning tasks for its business partner data . Therefore, the Orchestrator should offer an endpoint which accepts generic business partner data and returns a cleaning task identifier for the created cleaning task.

### Tasks
- [x] Endpoint definition in orchestrator API module
- [x] Orchestrator application module implementation with static dummy data return

Proposal for Endpoint:
[Redoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/eclipse-tractusx/bpdm/docs/api/generic-input/docs/api/orchestrator-proposal.yaml)

Dependencies:
#424 
#377

PR #481, opened 2023-09-21
Feat: Use interfaces for Generic Business Partner

martinfkaeser:
Use interfaces for the Generic Business Partner and its Postal Address.
Move them and dependent base classes to `bpdm-common` module.

Relates to #476 and #425


rainer-exxcellent (commented):
For me it looks good

bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/IBaseBusinessPartnerDto.kt
```diff
@@ -17,11 +17,10 @@
  * SPDX-License-Identifier: Apache-2.0
  ******************************************************************************/
 
-package org.eclipse.tractusx.bpdm.gate.api.model
+package org.eclipse.tractusx.bpdm.common.dto
 
 import io.swagger.v3.oas.annotations.media.ArraySchema
 import io.swagger.v3.oas.annotations.media.Schema
-import org.eclipse.tractusx.bpdm.common.dto.ClassificationDto
 import org.eclipse.tractusx.bpdm.common.dto.openapidescription.CommonDescription
```
nicoprow, line 25: Contains externalId Field. I think the externalId should be only part of the Gate and not be put in the common

nicoprow (changes requested):
Some consideration about the externalId field of the business partner base class in the common. See comment

bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/IBaseBusinessPartnerDto.kt
martinfkaeser: Because there is no `externalId` in the Orchstrator model? Done!

nicoprow: Yep, externalId is an identifier unique to the Gate. The orchestrator has cleaning task identifiers which the Gate uses to identify its own business partners

martinfkaeser:
Added `IBaseStreetDto` and `IStreetDetailedDto` to `bpdm-common` for better reusability.

nicoprow (approved):

PR #483, opened 2023-09-21
Refactor: Simplify Gate/Pool API clients

martinfkaeser:
Directly expose lazy properties, not functions, for the different APIs.
This is more straight-forward in Kotlin and simplifies the API client implementations.

### DIFF
--- a/bpdm-gate-api/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/api/client/GateClientImpl.kt
+++ b/bpdm-gate-api/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/api/client/GateClientImpl.kt
@@ -26,6 +26,12 @@ import org.springframework.web.reactive.function.client.support.WebClientAdapter
 import org.springframework.web.service.invoker.HttpServiceProxyFactory
 import java.time.Duration
 
+/**
+ * In a Spring configuration a bean of this class is instantiated passing a webClientProvider which configures the web client with e.g. OIDC configuration.
+ * A lazy HttpServiceProxyFactory private property is defined: On first access it creates a HttpServiceProxyFactory configured with the web client.
+ * Several lazy API clients are defined: On first access they are created from the HttpServiceProxyFactory for the specific API interface.
+ * All this has to be done lazily because during integration tests the web client URL may not be available yet on Spring initialization.
+ */
 class GateClientImpl(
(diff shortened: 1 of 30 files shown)


bpdm-gate-api/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/api/client/GateClientImpl.kt
```diff
@@ -38,26 +38,18 @@ class GateClientImpl(
...
-    override fun sites() = gateClientSite
+    override val changelog by lazy { createClient<GateChangelogApi>() }
 
-    override fun changelog() = gateClientChangelog
-
-    override fun sharingState() = gateClientSharingState
+    override val sharingState by lazy { createClient<GateSharingStateApi>() }
 
+    private inline fun <reified T> createClient() =
+        httpServiceProxyFactory.createClient(T::class.java)
```
rainer-exxcellent, line 54: I think, tis line needs a bit of documentation. It not so clear, what the httpServiceProxyFactory does.
Is it neccessary to create the client on every call, or could we cache the result somewhere?

rainer-exxcellent (commented):
Looks good, I think the httpServiceProxyFactory should be better documented.

bpdm-gate-api/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/api/client/GateClientImpl.kt
martinfkaeser: A client is not created on every call because these properties, e.g. `businessParters`, are lazy properties which are initialized on first access using the lambda function ` { createClient<GateBusinessPartnerApi>() }`.
The only thing I changed here is removing the unnecessary functions and moving the client-creation into an internal helper function.
I agree, some documentation might clarify things, both the old and the new version.

> I think, tis line needs a bit of documentation. It not so clear, what the httpServiceProxyFactory does. Is it neccessary to create the client on every call, or could we cache the result somewhere?

martinfkaeser: Together with Rainer I improved the documentation.

nicoprow (approved):

Issue #458, opened 2023-09-13
Orchestrator API: Cleaning Steps Endpoints


nicoprow:
The BPDM Orchestrator application needs endpoints in which cleaning services can integrate against. The idea here is to provide cleaning services a way to reserve cleaning tasks from the orchestrator, perform the actual cleaning asynchronously and then post the cleaning result back to the orchestrator.

For this, the orchestrator should have two cleaning steps endpoints:
1. Task Reservation: With this endpoint the requestor can reserve a number of cleaning tasks for a given cleaning step. The endpoint should return the next number of cleaning tasks together with the business partner data in cleaning.
2. Task Results: With this endpoint the requestor can post cleaning result for previously reserved tasks for a given cleaning step. The endpoint should accept an array of results for cleaning tickets by task identifier.

### Tasks
- [x] Create endpoint definition for task reservation in orchestrator api module
- [x] Create endpoint definition for task results in orchestrator api module
- [x] Implement endpoint definitions in orchestrator module by returning dummy data

[OpenApi Proposal](https://raw.githubusercontent.com/eclipse-tractusx/bpdm/docs/api/generic-input/docs/api/orchestrator-proposal.yaml) and the [Redoc Rendering](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/eclipse-tractusx/bpdm/docs/api/generic-input/docs/api/orchestrator-proposal.yaml)

nicoprow:
I set the type of the legal entity and site address fields to addressPartner. This way the cleaning service can specify the full legal and site main addresses including bpn-reference and identifiers. I also renamed the address fields to legal address and site main address respectively.

PR #498, opened 2023-09-27
Feat: Add Endpoint for Posting Cleaning Results to Orchestrator

nicoprow:
## Description

This pull request introduces an endpoint to post cleaning results for previously reserved cleaning tasks to the orchestrator. The API endpoint is implemented in the orchestrator controller and returns dummy data on invocation.

I also added tests checking the upsert limit and whether the endpoint can be invoked.

For the necessary test I created template objects for L/S/A business partners that can be used by the tests.

Solves issue #458

## Pre-review checks

### DIFF
--- a/bpdm-orchestrator-api/src/main/kotlin/org/eclipse/tractusx/orchestrator/api/CleaningTaskApi.kt
+++ b/bpdm-orchestrator-api/src/main/kotlin/org/eclipse/tractusx/orchestrator/api/CleaningTaskApi.kt
@@ -100,4 +100,26 @@ interface CleaningTaskApi {
     @PostMapping("/reservations")
     @PostExchange("/reservations")
     fun reserveCleaningTasks(@RequestBody reservationRequest: CleaningReservationRequest): CleaningReservationResponse
+
+    @Operation(
+        summary = "Post cleaning results for reserved cleaning tasks in given cleaning step",
+        description = "Post business partner cleaning results for the given cleaning tasks. " +
+                "In order to post a result for a cleaning task it needs to be reserved first, has to currently be in the given cleaning step and the time limit is not exceeded." +
+                "The number of results you can post at a time does not need to match the original number of reserved tasks." +
+                "Results are accepted via strategy 'all or nothing'." +
+                "For a single request, the maximum number of postable results is limited to \${bpdm.api.upsert-limit}."
+    )
+    @ApiResponses(
+        value = [
+            ApiResponse(
+                responseCode = "204",
+                description = "If the results could be processed"
+            ),
+            ApiResponse(responseCode = "400", description = "On malformed task create requests or reaching upsert limit", content = [Content()]),
+        ]
+    )
+    @Tag(name = "Cleaning Service")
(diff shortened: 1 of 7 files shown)


nicoprow:
Rebased on top of main and ready to be reviewed

bpdm-orchestrator-api/src/main/kotlin/org/eclipse/tractusx/orchestrator/api/CleaningTaskApi.kt
```diff
@@ -97,4 +103,26 @@ interface CleaningTaskApi {
...
+        description = "Post business partner cleaning results for the given cleaning tasks. " +
+                "In order to post a result for a cleaning task it needs to be reserved first, has to currently be in the given cleaning step and the time limit is not exceeded." +
+                "The number of results you can post at a time does not need to match the original number of reserved tasks." +
+                "Results are accepted via strategy 'all or nothing'." +
+                "For a single request, the maximum number of postable results is limited to \${bpdm.api.upsert-limit}."
+    )
+    @ApiResponses(
+        value = [
+            ApiResponse(
+                responseCode = "200",
```
martinfkaeser, line 118: To be precise, 204 would be the correct response code

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/controller/CleaningTaskController.kt
```diff
@@ -52,6 +52,11 @@ class CleaningTaskController(
         }
     }
 
+    override fun resolveCleaningTasks(resultRequest: CleaningResultRequest) {
+        if (resultRequest.results.size > apiConfigProperties.upsertLimit)
+            throw BpdmUpsertLimitException(resultRequest.results.size, apiConfigProperties.upsertLimit)
+    }
```
martinfkaeser, line 58: I suggest we could also check if at least result or any errors exist.
```suggestion
        resultRequest.results.forEach { entry -> 
            if (entry.result == null && entry.errors.isEmpty())
                throw Bpdm...
        }
```
And we could add a test for that.

martinfkaeser:
Besides my comments everythink looks good!

bpdm-orchestrator-api/src/main/kotlin/org/eclipse/tractusx/orchestrator/api/CleaningTaskApi.kt
nicoprow: Fixed it

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/controller/CleaningTaskController.kt
nicoprow: I added this validation logic but I was a bit torn about it as I think it could also be part of a separate validation service logic and not in the controller. We can still move it later if necessary

Issue #499, opened 2023-09-28
Pool-API: Create administrative Level 1 endpoint


nicoprow:
Currently, in our data model the postal addresses contain a field for setting the administrative level 1 value. We expect here an ISO norm region code to be entered. For this, we created earlier metadata endpoints in the Pool which have the path "api/catena/regions". For a user it is confusing to have two different terminologies for the same concept, on time being called administrative level 1 and at another place called region. 

Therefore we should align the naming by introducing a new Pool Metadata endpoint and deprecating the old ones.

### Tasks
- [x] Add Metadata GET Endpoint "api/catena/administrative-areas-level1" (POST is no longer needed)
- [x] Connect the new GET endpoint with the existing service logic and region DTOs
- [x] Mark the region endpoints as deprecated

nicoprow:
We should rename regions in the new endpoint with administrative level 1. 
Also:
regionCode -> code 
regionName -> name

PR #502, opened 2023-09-28
Feat: Create administrative areas level 1 endpoint

martinfkaeser:
Create api/catena/administrative-areas-level1 (GET) endpoint as replacement for the confusingly named api/catena/regions (GET and POST) endpoints. The later two endpoints are marked as deprectated in OpenAPI.
The endpoint sorts the returned regions by `countryCode` and `regionCode`.

Solves #499

### DIFF
--- a/bpdm-pool/src/test/kotlin/org/eclipse/tractusx/bpdm/pool/controller/MetadataControllerIT.kt
+++ b/bpdm-pool/src/test/kotlin/org/eclipse/tractusx/bpdm/pool/controller/MetadataControllerIT.kt
@@ -61,7 +61,7 @@ private typealias GetFunction = (client: WebTestClient, page: Int, size: Int) ->
     webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = [Application::class, TestHelpers::class]
 )
 @ActiveProfiles("test")
-@ContextConfiguration(initializers = [PostgreSQLContextInitializer::class,OpenSearchContextInitializer::class])
+@ContextConfiguration(initializers = [PostgreSQLContextInitializer::class, OpenSearchContextInitializer::class])
 class MetadataControllerIT @Autowired constructor(
     private val testHelpers: TestHelpers,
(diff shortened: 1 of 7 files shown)


bpdm-pool/src/test/kotlin/org/eclipse/tractusx/bpdm/pool/controller/MetadataControllerIT.kt
```diff
@@ -425,4 +432,30 @@ class MetadataControllerIT @Autowired constructor(
...
+        val page = poolClient.metadata.getAdminAreasLevel1(PaginationRequest())
+        assertThat(page.totalElements).isEqualTo(7L)
+        assertThat(page.content.size).isEqualTo(7)
+        val (countryCode, regionCode, regionName) = page.content.first { it.regionCode == "DE-BE" }
+        assertThat(countryCode).isEqualTo(DE)
+        assertThat(regionCode).isEqualTo("DE-BE")
+        assertThat(regionName).isEqualTo("Berlin")
+    }
+
+    private fun initAdminAreasLevel1() {
```
nicoprow, line 448: I just noticed that when we clear our database between tests we also lose the regions we gained through the database migration. In my opinion the should assume that the region values are always in the database as they are based on an ISO norm. So it would be better exclude the table 'regions' from the database wipe and adapt the test to check that the endpoint returns the first page of iso norm regions

nicoprow (changes requested):
Made a proposal for testing behaviour

bpdm-pool/src/test/kotlin/org/eclipse/tractusx/bpdm/pool/controller/MetadataControllerIT.kt
martinfkaeser: Good idea!
This made me thinking that the regions should better be returned in a predictable order.
Now they are sorted by `countryCode` and `regionCode`.

nicoprow (approved):

PR #503, opened 2023-10-02
 refactor(gate): business partner service

nicoprow:
## Description

This pull request is a proposal on how services employ business logic to DTOs and entities. Currently, often a service takes a DTO and employs business logic on it, especially in the case of updates or inserts. The problem here is that a lot of this business logic can not be reused and need to be duplicated if a new DTO is introduced to the service. 

There is the option to convert from DTO to DTO but this establishes more dependencies between the DTOs as well as dependency between service and the DTO which should actually abstract from the model. Instead, I propose for the services to use the actual model classes as early as possible. Only applying logic for DTO-specific validation before converting the DTO to its model representation and then employ the business logic from there.

In the pull request itself I used the BusinessPartnerService as an example because it clearly demonstrates that a lot of business logic is applied to the generic business partner DTO even though we want to reuse the service logic for the L/S/A business partner DTOs as well. By introducing the refactorings in the BusinessPartnerService we are now able to reuse most of the business logic with both DTO models.

## Pre-review checks

### DIFF
--- a/bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
+++ b/bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
@@ -19,22 +19,24 @@
 
 package org.eclipse.tractusx.bpdm.gate.service
 
-import org.eclipse.tractusx.bpdm.common.dto.AddressType
 import org.eclipse.tractusx.bpdm.common.dto.BusinessPartnerType
 import org.eclipse.tractusx.bpdm.common.dto.response.PageDto
 import org.eclipse.tractusx.bpdm.common.model.StageType
 import org.eclipse.tractusx.bpdm.common.service.toPageDto
+import org.eclipse.tractusx.bpdm.common.util.copyAndSync
+import org.eclipse.tractusx.bpdm.common.util.replace
 import org.eclipse.tractusx.bpdm.gate.api.model.ChangelogType
-import org.eclipse.tractusx.bpdm.gate.api.model.IBaseBusinessPartnerGateDto
 import org.eclipse.tractusx.bpdm.gate.api.model.request.BusinessPartnerInputRequest
 import org.eclipse.tractusx.bpdm.gate.api.model.request.BusinessPartnerOutputRequest
 import org.eclipse.tractusx.bpdm.gate.api.model.response.BusinessPartnerInputDto
 import org.eclipse.tractusx.bpdm.gate.api.model.response.BusinessPartnerOutputDto
 import org.eclipse.tractusx.bpdm.gate.api.model.response.SharingStateDto
 import org.eclipse.tractusx.bpdm.gate.entity.ChangelogEntry
-import org.eclipse.tractusx.bpdm.gate.entity.generic.BusinessPartner
+import org.eclipse.tractusx.bpdm.gate.entity.generic.*
+import org.eclipse.tractusx.bpdm.gate.exception.BpdmMissingStageException
 import org.eclipse.tractusx.bpdm.gate.repository.ChangelogRepository
 import org.eclipse.tractusx.bpdm.gate.repository.generic.BusinessPartnerRepository
+import org.springframework.data.domain.Page
 import org.springframework.data.domain.PageRequest
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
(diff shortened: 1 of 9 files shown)


fabiodmota (approved):
LGTM  the logic seems well divided between what is INPUT and the OUTPUT flow.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -49,110 +50,137 @@ class BusinessPartnerService(
...
+        return upsertBusinessPartnersInput(entities).map(businessPartnerMappings::toBusinessPartnerInputDto)
+    }
 
-        val savedEntities = dtos.map { dto ->
-            existingEntitiesByExternalId[dto.externalId]
-                ?.let { existingEntity -> updateBusinessPartnerInput(existingEntity, dto) }
-                ?: run { insertBusinessPartnerInput(dto) }
-        }
+    @Transactional
+    fun upsertBusinessPartnersOutput(dtos: Collection<BusinessPartnerOutputRequest>): Collection<BusinessPartnerInputDto> {
```
martinfkaeser, line 58: Return type should be `Collection<BusinessPartnerOutputDto>`

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -49,110 +50,137 @@ class BusinessPartnerService(
...
+        toPartner.identifiers = fromPartner.identifiers
+        toPartner.states = fromPartner.states
+        toPartner.classifications = fromPartner.classifications
+        toPartner.shortName = fromPartner.shortName
+        toPartner.legalForm = fromPartner.legalForm
+        toPartner.isOwner = fromPartner.isOwner
+        toPartner.bpnL = fromPartner.bpnL
+        toPartner.bpnS = fromPartner.bpnS
+        toPartner.bpnA = fromPartner.bpnA
+        toPartner.postalAddress = fromPartner.postalAddress
```
martinfkaeser, line 115: I'm not sure if we want to replace the whole `PostalAddress` with a new one. At least orphans are removed via annotation.
This works, but it's different to the way we handle `BusinessPartner`s. Theoretically we could also delete them and create new ones on update.
Alternative would be a `copyValues` for `PostalAddress`.
What do you think is preferable?

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -49,110 +50,137 @@ class BusinessPartnerService(
...
-        val page = when {
-            externalIds.isNullOrEmpty() -> businessPartnerRepository.findByStage(stage, pageRequest)
-            else -> businessPartnerRepository.findByStageAndExternalIdIn(stage, externalIds, pageRequest)
+    private fun assertInputStageExists(externalIds: Collection<String>) {
+        val existingExternalIds = businessPartnerRepository.findByStageAndExternalIdIn(StageType.Input, externalIds)
+            .map { it.externalId }
+            .toSet()
+
+        if (externalIds.minus(existingExternalIds).isNotEmpty()) {
+            val missing = externalIds.minus(existingExternalIds)
```
martinfkaeser, line 155: You could move the variable out of the if block and reuse it.
Better naming: `missingExternalIds` ?

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -49,110 +50,137 @@ class BusinessPartnerService(
...
+            throw BpdmBusinessPartnerStageNotExistsException(missing.first(), StageType.Input)
```
martinfkaeser, line 156: Better naming maybe: 
`BpdmBusinessPartnerMissingStageException`

martinfkaeser (changes requested):

martinfkaeser:
I like that majority of methods in `BusinessPartnerService` now deal with `BusinessPartner` and can be better reused.
Also I like that we can get rid of the awkward `updateBusinessPartner` methods in `BusinessPartnerMappings` which didn't really fit.

alexsilva-CGI:
Looks good from me as well. The BusinessPartnerService using BusinessPartner can will definitely help with the mappings of the L/S/A Types by mapping them cleaner, and parent relation as well.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -49,110 +50,137 @@ class BusinessPartnerService(
...
-    private fun initSharingState(dto: IBaseBusinessPartnerGateDto) {
-        // TODO make businessPartnerType optional
-        sharingStateService.upsertSharingState(SharingStateDto(BusinessPartnerType.ADDRESS, dto.externalId))
+        val newEntities = mutableListOf<BusinessPartner>()
+        val existingEntities = mutableListOf<BusinessPartner>()
+
+        entityCandidates.forEach { candidate ->
+            val existingEntity = existingPartnersByExternalId[candidate.externalId]
+            if (existingEntity != null)
+                existingEntities.add(copyValues(candidate, existingEntity))
```
rainer-exxcellent, line 171: Could we do the saveChangelog here? Then it is maybe not neccessary to split in new and existing entities.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -49,110 +50,137 @@ class BusinessPartnerService(
...
+        toPartner.bpnS = fromPartner.bpnS
+        toPartner.bpnA = fromPartner.bpnA
+        toPartner.postalAddress = fromPartner.postalAddress
+        toPartner.parentId = fromPartner.parentId
+        toPartner.parentType = fromPartner.parentType
+
+        return toPartner
     }
 
-    private fun checkBusinessPartnerType(type: AddressType?): List<BusinessPartnerType>? {
```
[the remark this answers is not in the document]
martinfkaeser: I kind of like it in this method. I don't know if we want the tighter coupling in the enum class.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
nicoprow: I decided to add copyValues methods for all independent entities attached to business partner entity

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
nicoprow: Right, I adjusted the naming

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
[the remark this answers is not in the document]
nicoprow: We decided to add the business partner types to the AddressType enum for now. They reside both in the common library and it seems this information will be handy in more places than just the business partner service.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
nicoprow: Done

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
nicoprow: I reworked this and now create a new internal data class for entity resolution results and this is then used for creating the changelogs

PR #511, opened 2023-10-04
Feat: Administrative Level 1 endpoint - don't call it "region" in descriptions and properties

martinfkaeser:
For endpoint and DTOs rephrase *region* to *administrativeAreaLevel1* and *country subdivision* because region is not an official term according to ISO 3166-2.

Solves #499

### DIFF
--- a/bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/response/CountrySubdivisionDto.kt
+++ b/bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/response/CountrySubdivisionDto.kt
@@ -22,15 +22,15 @@ package org.eclipse.tractusx.bpdm.common.dto.response
 import com.neovisionaries.i18n.CountryCode
 import io.swagger.v3.oas.annotations.media.Schema
 
-@Schema(description = "Region within a country")
-data class RegionDto(
+@Schema(description = "Country subdivision")
+data class CountrySubdivisionDto(
 
     @get:Schema(description = "Country code")
     val countryCode: CountryCode,
 
-    @get:Schema(description = "Abbreviation or shorthand of the area")
-    val regionCode: String,
+    @get:Schema(description = "The country subdivision code according to ISO 3166-2")
+    val code: String,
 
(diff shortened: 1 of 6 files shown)


bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/response/RegionDto.kt
```diff
@@ -22,15 +22,15 @@ package org.eclipse.tractusx.bpdm.common.dto.response
...
 
-    @get:Schema(description = "Abbreviation or shorthand of the area")
-    val regionCode: String,
+    @get:Schema(description = "The country subdivision code according to ISO 3166-2")
+    val code: String,
 
-    @get:Schema(description = "Describes the full name of the region within a country according to ISO 3166-214")
-    val regionName: String
+    @get:Schema(description = "The name of the country subdivision according to ISO 3166-2")
+    val name: String
```
nicoprow, line 35: We need to have another DTO since we still need to old Region endpoints to be backwards compatible

nicoprow (changes requested):

bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/dto/response/RegionDto.kt
martinfkaeser: Fixed it. Now there is a new `CountrySubdivisionDto` besides the old `RegionDto`.

nicoprow (approved):

nicoprow (changes requested):
Bridge fails to compile with the new changes

martinfkaeser:
> Bridge fails to compile with the new changes

The problem is that the Pool-API returns a `RegionDto` inside its `PhysicalPostalAddressVerboseDto`/ `AlternativePostalAddressVerboseDto`.
Now the question is: Do we change it to `CountrySubdivisionDto` which seems like the reasonable thing but breaks the API? Or do we keep the deprecated `RegionDto` inside both DTOs?

--> After talking with @nicoprow we decided to keep the `RegionDto` inside the address DTOs.

Issue #427, opened 2023-09-01
Orchestrator: Manage Cleaning Request State and Data


nicoprow:
The BPDM Orchestrator should manage the life cycle of a cleaning request. By this it needs to implement a state machine, describing logic when the cleaning request moves from one state to another.

```mermaid
---
title: Cleaning Task Transition Logic
---
stateDiagram-v2
    
    state "Queued" as GenericCleaningQueued
    state "Reserved" as GenericCleaningReserved

    state "Queued" as BpnProcessingQueued
    state "Reserved" as BpnProcessingReserved

    state "Queued" as GoldenRecordUpdateQueued
    state "Reserved" as GoldenRecordUpdateReserved

    [*] --> Pending

    state Pending{
         
         [*] --> GenericCleaning: Mode - UpdateFromSharingMember

        state GenericCleaning{
            [*] --> GenericCleaningQueued
            GenericCleaningQueued --> GenericCleaningReserved
            GenericCleaningReserved --> [*]
        }

        GenericCleaning --> BpnProcessing

        state BpnProcessing{
            [*] --> BpnProcessingQueued
            BpnProcessingQueued --> BpnProcessingReserved
            BpnProcessingReserved --> [*]
        }

        BpnProcessing --> [*]

        [*] --> GoldenRecordUpdate: Mode - UpdateFromPool
        
        state GoldenRecordUpdate{
            [*] --> GoldenRecordUpdateQueued
            GoldenRecordUpdateQueued --> GoldenRecordUpdateReserved
            GoldenRecordUpdateReserved --> [*]
        }

        GoldenRecordUpdate --> [*]

    }

    Pending --> Success: if finished
    Pending --> Error: if disrupted by error

    Success --> [*]
    Error --> [*]

```

```mermaid

---
title: Cleaning Task Composition
---
classDiagram
    
   class CleaningTask{
        Task Identifier: String
        Generic Business Partner: Object
        Legal Entity: Object
        Site: Object
        Address: Object
   }

   class TaskState{
        Result State: ResultStateEnum
        CreatedAt: Timestamp
        ModifiedAt: Timestamp 
   }

   class ResultState{
        <<enum>>
        Pending
        Success
        Error
   }

   class StepState{
        CreatedAt: Timestamp
   }

   class CleaningStep{
        <<enum>>
        GenericCleaning
        BpnProcessing
        GoldenRecordUpdate
   }

   class ReservationState{
        <<enum>>
        Queued
        Reserved
[...]

PR #517, opened 2023-10-12
Orchestrator: Service logic

martinfkaeser:
The Orchestrator endpoints have been properly implemented replacing the dummy implementation. The Orchestrator now manages the life cycle of Golden Record tasks. 

The Orchestrator API has been slightly modified following the discussion of https://github.com/eclipse-tractusx/bpdm/issues/427#issuecomment-1754905430
- `StepState` has been extended
- `timeout` has been added to `TaskProcessingStateDto`

This PR contains unit tests for the GoldenRecordTaskStateMachine and endpoint integration tests.

Solves #427


bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/GoldenRecordTaskStorage.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.eclipse.tractusx.bpdm.orchestrator.model.GoldenRecordTask
+import org.eclipse.tractusx.orchestrator.api.model.ResultState
+import org.eclipse.tractusx.orchestrator.api.model.StepState
+import org.eclipse.tractusx.orchestrator.api.model.TaskStep
+import org.springframework.stereotype.Service
+
+@Service
+class GoldenRecordTaskStorage {
+
+    private val tasks: MutableList<GoldenRecordTask> = mutableListOf()
```
nicoprow, line 32: This may lead to concurrency exceptions if multiple threads from different requests access the list at the same time. I think the long term solution is to store it in a database anyway, so I think we should just think of a very low effort solution right now. Simpliest solution that I can think of right now is to limit the request thread pool to 1 basically ensuring sequential processing of requests: [server.tomcat.max-connections](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#application-properties.server.server.tomcat.max-connections)

nicoprow (changes requested):

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/GoldenRecordTaskStorage.kt
martinfkaeser: Or we could synchronize (`@Synchronize`) all the methods in `GoldenRecordTaskService`?
It's basically the same but is a little more fine grained and explicit.
This shouldn't have much impact as these methods are expected to be very fast.

nicoprow: Yea that should work as well for now

nicoprow (approved):

nicoprow (approved):

Issue #432, opened 2023-09-01
Pool: Upsert Business Partners from Cleaning Result


nicoprow:
In the new orchestration logic the BPDM Pool should poll the BPDM Orchestrator for new cleaning requests that need to be processed. The Pool processes the cleaning requests first by upserting its own business partners via the attached L/S/A cleaning result. The cleaning result may have BPN request identifiers which the Pool needs to convert to BPNs with the following logic:

* If the BPN request identifier is unknown create a new business partner, generate a new BPN and associate the unknown identifier with the generated BPN. For the L/S/A hierarchy:If there is a site attach the address to site, otherwise attach the address to the legal entity.
* If the BPN request identifier is known update the business partner with the associated BPN. Fow now, updates do **not** change the L/S/A hierarchy

By this, the Pool upserts business partners from a given cleaning result.
In a final step, the Pool should replace all BPN request identifiers with the associated BPNs in the cleaning result and send that cleaning result to the orchestrator for further processing.

###Tasks
- [x] Create polling logic for reserving cleaning task for step "BpnProcessing" from Orchestrator (see #458) 
- [x] Create BPN upsert logic based on BPN request identifier
- [x] Create upsert logic from L/S/A business partners in cleaning tasks
- [x] Create cleaning result and post results back for step "BpnProcessing" to Orchestrator (see #458)

Dependencies:
#435 
#434

nicoprow:
I changed the type of addresses for the legal entity and site cleaning results to AddressPartner. With that the address fields now contain bpna-References and identifiers. With that the cleaning service can specify the full legal address and site main address.

PR #518, opened 2023-10-12
Feat: Upsert Pool from Cleaning Result - Refactor validation

rainer-exxcellent:
## Description
- create initial poll logic
- refactor the validation for the creation of legal entities so that it could be used by the orchestrator DTOs

Implements #432  partly

### DIFF
--- a/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/FetchAndReserveCleaningStepService.kt
+++ b/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/FetchAndReserveCleaningStepService.kt
@@ -1,0 +1,204 @@
+/*******************************************************************************
+ * Copyright (c) 2021,2023 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License, Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ******************************************************************************/
+
+package org.eclipse.tractusx.bpdm.pool.service
+
+import mu.KotlinLogging
+import org.eclipse.tractusx.bpdm.common.dto.IBaseLegalEntityDto
+import org.eclipse.tractusx.bpdm.pool.api.model.response.ErrorInfo
+import org.eclipse.tractusx.bpdm.pool.api.model.response.LegalEntityCreateError
+import org.eclipse.tractusx.bpdm.pool.dto.LegalEntityMetadataDto
+import org.eclipse.tractusx.bpdm.pool.entity.*
+import org.eclipse.tractusx.bpdm.pool.entity.LegalEntityState
+import org.eclipse.tractusx.orchestrator.api.client.OrchestrationApiClient
+import org.eclipse.tractusx.orchestrator.api.model.*
+import org.springframework.scheduling.annotation.Scheduled
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+
+class FetchAndReserveCleaningStepService(
+    private val orchestrationClient: OrchestrationApiClient,
+    private val metadataService: MetadataService,
+    private val requestValidationService: RequestValidationService,
+    private val bpnIssuingService: BpnIssuingService,
+) {
+    private val logger = KotlinLogging.logger { }
+
+    @Scheduled(cron = "\${bpdm.opensearch.export-scheduler-cron-expr:-}", zone = "UTC")
+    fun fetchAndReserve() {
+
+        val reservationRequest = TaskStepReservationRequest(step = TaskStep.PoolSync, amount = 10)
+        val taskStepReservation = orchestrationClient.goldenRecordTasks.reserveTasksForStep(reservationRequest = reservationRequest)
+
+        val taskResults = upsertGoldenRecordIntoPool(taskStepReservation.reservedTasks)
+        orchestrationClient.goldenRecordTasks.resolveStepResults(TaskStepResultRequest(step=TaskStep.PoolSync,results = taskResults))
+    }
+
+    private fun upsertGoldenRecordIntoPool(tasks: List<TaskStepReservationEntryDto>): List<TaskStepResultEntryDto> {
+
+        val isTaskCreateLegalEntity =
+            { task: TaskStepReservationEntryDto -> task.businessPartner.legalEntity?.bpnLReference?.referenceType == BpnReferenceType.BpnRequestIdentifier }
+
+        val legalEntitiesToCreateSteps: List<TaskStepReservationEntryDto> = tasks
+            .filter { isTaskCreateLegalEntity(it) }
+
+        val legalEntityByTask = legalEntitiesToCreateSteps
+            .associateWith { it.businessPartner.legalEntity as IBaseLegalEntityDto }
+
+        val addressByTask = legalEntitiesToCreateSteps
+            .associateWith { it.businessPartner.legalEntity?.legalAddress as LogisticAddressDto }
+
+
+        val errorsByRequest = requestValidationService.validateLegalEntityCreatesOrchestrator(legalEntityByTask)
+        val errorsByRequestAddress = requestValidationService.validateLegalEntityCreatesAddressesOrchestrator(addressByTask)
+
+        val legalEntityCreateTaskResults = legalEntitiesToCreateSteps
+            .map { taskStep ->
+                if (errorsByRequest.containsKey(taskStep) || errorsByRequestAddress.containsKey(taskStep)) {
+                    taskResultsForErrors(
+                        taskStep.taskId,
+                        errorsByRequest.getOrDefault(taskStep, emptyList()) + errorsByRequestAddress.getOrDefault(taskStep, emptyList())
+                    )
+                } else {
+                    legalEntityCreateTaskResult(taskStep)
+                }
+            }
+        return legalEntityCreateTaskResults
+    }
+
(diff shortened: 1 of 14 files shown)


bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/RequestValidationService.kt
```diff
@@ -40,56 +38,76 @@ class RequestValidationService(
...
 ) {
-    fun validateLegalEntityCreates(
-        requests: Collection<LegalEntityPartnerCreateRequest>
-    ): Map<LegalEntityPartnerCreateRequest, Collection<ErrorInfo<LegalEntityCreateError>>> {
 
-        val legalEntityRequests = requests.map { it.legalEntity }
-        val legalAddressRequests = requests.map { it.legalAddress }
+    fun <IRequest>validateLegalEntityCreates(
+        legalEntitiesByRequest: Map<IRequest, IBaseLegalEntityDto>, entityKeyFunc : (IRequest) -> String?
+    ): Map<IRequest, List<ErrorInfo<LegalEntityCreateError>>> {
```
nicoprow, line 44: This is quite a complicated function call for being used by other services. Maybe make it private and add public functions that just accept the actual DTOs of Pool and Orchestrator? Then the invoking services have an easier time using this service

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/FetchAndReserveCleaningStepService.kt
```diff
@@ -0,0 +1,198 @@
...
+
+class FetchAndReserveCleaningStepService(
+    private val orchestrationClient: OrchestrationApiClient,
+    private val metadataService: MetadataService,
+    private val requestValidationService: RequestValidationService,
+    private val bpnIssuingService: BpnIssuingService,
+) {
+    private val logger = KotlinLogging.logger { }
+
+    @Scheduled(cron = "\${bpdm.opensearch.export-scheduler-cron-expr:-}", zone = "UTC")
```
martinfkaeser, line 44: This is the OpenSearch cron expression. We need a separate one for the Orchestrator.

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/FetchAndReserveCleaningStepService.kt
```diff
@@ -0,0 +1,198 @@
...
+import org.eclipse.tractusx.bpdm.pool.dto.LegalEntityMetadataDto
+import org.eclipse.tractusx.bpdm.pool.entity.*
+import org.eclipse.tractusx.bpdm.pool.entity.LegalEntityState
+import org.eclipse.tractusx.orchestrator.api.client.OrchestrationApiClient
+import org.eclipse.tractusx.orchestrator.api.model.*
+import org.springframework.scheduling.annotation.Scheduled
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+
+class FetchAndReserveCleaningStepService(
```
martinfkaeser, line 36: `@Service` needed

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/FetchAndReserveCleaningStepService.kt
```diff
@@ -0,0 +1,198 @@
...
+) {
+    private val logger = KotlinLogging.logger { }
+
+    @Scheduled(cron = "\${bpdm.opensearch.export-scheduler-cron-expr:-}", zone = "UTC")
+    fun fetchAndReserve() {
+
+        val reservationRequest = TaskStepReservationRequest(step = TaskStep.PoolSync, amount = 10)
+        val taskStepReservation = orchestrationClient.goldenRecordTasks.reserveTasksForStep(reservationRequest = reservationRequest)
+
+        val taksResults = upsertGoldenRecordIntoPool(taskStepReservation.reservedTasks)
```
martinfkaeser, line 50: typo

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/FetchAndReserveCleaningStepService.kt
```diff
@@ -0,0 +1,198 @@
...
+    private fun upsertGoldenRecordIntoPool(tasks: List<TaskStepReservationEntryDto>): List<TaskStepResultEntryDto> {
+
+        val isTaskCreateLegalEntity = { task: TaskStepReservationEntryDto -> task.businessPartner.legalEntity?.bpnLReference?.referenceType == BpnReferenceType.BpnRequestIdentifier }
+
+        val legalEntitiesToCreateSteps : List<TaskStepReservationEntryDto> = tasks
+            .filter { isTaskCreateLegalEntity(it) }
+
+        val legalEntityByTask = legalEntitiesToCreateSteps
+            .associateWith{ it.businessPartner.legalEntity as IBaseLegalEntityDto}
+            .toMap()
```
martinfkaeser, line 63: `toMap` is redundant

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/FetchAndReserveCleaningStepService.kt
```diff
@@ -0,0 +1,198 @@
...
+            .associateWith{ it.businessPartner.legalEntity?.legalAddress as IBaseLogisticAddressDto}
+            .toMap()
+
+        val errorsByRequest = requestValidationService.validateLegalEntityCreates(legalEntityByTask) { task -> task.businessPartner.legalEntity?.bpnLReference?.referenceValue }
+        val errorsByRequestAddress = requestValidationService.validateLegalEntityCreatesAddresses(addressByTask) { task -> task.businessPartner.legalEntity?.bpnLReference?.referenceValue }
+
+        val legalEntityCreateTaskResults = legalEntitiesToCreateSteps
+            .map{ taskStep ->
+                if (errorsByRequest.containsKey(taskStep) || errorsByRequestAddress.containsKey(taskStep)) {
+                    taskResultsForErrors(taskStep.taskId,errorsByRequest.getOrDefault(taskStep, mutableListOf()) +  errorsByRequestAddress.getOrDefault(taskStep, mutableListOf()) )
```
martinfkaeser, line 74: You don't need a `mutableListOf()`, you can use `emptyList()` instead.

martinfkaeser (changes requested):

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/FetchAndReserveCleaningStepService.kt
```diff
@@ -0,0 +1,198 @@
...
+
+        legalEntity.currentness = createCurrentnessTimestamp()
+
+        legalEntity.legalName = legalName
+
+        legalEntity.identifiers.clear()
+        legalEntity.states.clear()
+        legalEntity.classifications.clear()
+
+        legalEntity.states.addAll(request.states.map { toEntity(it, legalEntity) })
```
martinfkaeser, line 152: There is a `.replace(..)` helper function that does `.clear()` and `.addAll()` in together.

martinfkaeser (changes requested):

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/FetchAndReserveCleaningStepService.kt
rainer-exxcellent: This is done in the next Step. Focus on this pull request is on refactoring

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/FetchAndReserveCleaningStepService.kt
rainer-exxcellent: the whole method is refactored in the next step

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/RequestValidationService.kt
rainer-exxcellent: The split is  definitely important. One Method should do on task. The logic for addresses and legal entities are completely independent from each other.
The validation logic is complicated, so i think should be reused. If we copy the code at the first level it would be easier to understand.

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/MetadataService.kt
```diff
@@ -152,18 +152,35 @@ class MetadataService(
...
-            .plus(requests.mapNotNull { it.alternativePostalAddress?.administrativeAreaLevel1 })
+        val regionKeys = requests.mapNotNull { administrativeAreaLevel1ToString(it.physicalPostalAddress?.administrativeAreaLevel1) }
+            .plus(requests.mapNotNull { administrativeAreaLevel1ToString(it.alternativePostalAddress?.administrativeAreaLevel1) })
             .toSet()
         val regions = regionRepository.findByRegionCodeIn(regionKeys)
 
         return AddressMetadataDto(idTypes, regions)
     }
 
+    private fun administrativeAreaLevel1ToString(administrativeAreaLevel1: Any?): String? {
```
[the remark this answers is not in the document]
rainer-exxcellent: If we don't use data classes in this case, we could put this easily in a abstract method .
In this way we have to copy the code.

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/RequestValidationService.kt
martinfkaeser: We now have these 2 very related validation methods `validateLegalEntityCreates`, `validateLegalEntityCreatesAddresses`. First one is for legal entities, second for legal addresses beneath legal entities.
As far as I can see from a consumer side these two methods are always used together and the results are joined together (in FetchAndReserveCleaningStepService and BusinessPartnerBuildService). So I still think a single *public* method would be easier to use. But if it makes sense to have two separate methods because the logic is independent, you could make them private.

nicoprow (approved):

nicoprow:
These changes have already merged with pull reqeust #538

Issue #534, opened 2023-10-17
Orchestrator: Add Authentication/Authorization Logic


nicoprow:
Currently, the endpoints in the orchestrator application are not authenticated and not assigned any permissions. We should add authorization logic to the application analogous to the BPDM Gate and Pool. The endpoints should be secured with permissions having configurable names.

Suggestions:
- Task Client Controller - Create Task: create_task
- Task Client Controller - Search Task: view_task
- Task Worker Controller: process_task_step_STEPNAME, meaning cleanAndSync,_ process_task_step_poolSync, process_task_stepClean

martinfkaeser:
1) For the Task Worker role I would suggest naming it `process_task_step_STEPNAME` so we can use simple concatenation, e.g. `process_task_step_CleanAndSync`.

2) For the `resolveStepResults` function I suggest adding a `step: TaskStep` property to `TaskStepResultRequest`.
This would be an API change but it would greatly simplify the role check and make it more straightforward.
I think we can still adjust the API. The required changes are only minor.

@nicoprow

nicoprow:
Sure, sounds good to me. No problem to adjust the new Orchestor API if there is good reason for it. So I also think the step enum value would be beneficial here

PR #535, opened 2023-10-18
feat: Extend TaskStepResultRequest in orchestrator-api with step

martinfkaeser:
When posting task results (`POST /api/golden-record-tasks/step-results`) the request DTO now requires a `step` enum.
This is a precondition for authorization because permission should be checked based on step-based roles.

Relates to #534


nicoprow (changes requested):
Looks good but please rebase on the latest main as I can see problems arise with the new cleaning service dummy logic that is missing the step

nicoprow (approved):

Issue #452, opened 2023-09-12
CleaningServiceDummy: Create Dummy Golden Record Update


nicoprow:
We lack a service provider who can update a BPDM Gate generic business partner based on Pool L/S/A typed business partners. Idea for this dummy logic is to receive the generic business partner to clean and just send the same data back.  

### Tasks
- [x] Poll reservation endpoint for step "GoldenRecordUpdate" from Orchestrator (see #427)
- [x] Post result for step "GoldenRecordUpdate" to Orchestrator (see #427)

PR #544, opened 2023-10-19
Feat(Cleaning): Add logic for bpn cleaning upon pool update

fabiodmota:
## Description
Add logic for when schedule runs to trigger also clean flow 
With this we validate also when a bpn is updated from the pool 

https://github.com/eclipse-tractusx/bpdm/issues/452 

## Pre-review checks


nicoprow (approved):
Should work, even though we create a LSA business partner in the clean step, which is not required for that step actually. In any case it doesn't hurt the process

Issue #510, opened 2023-10-04
Gate: Rename Owner Field


nicoprow:
A better name for the "isOwner" field in the genric business partner of the Gate would be "IsOwnCompanyData".

### Tasks
- [x] Rename the field "isOwner" to "isOwnCompanyData"

nicoprow:
Field is still overwritten in OpenAPI Doc as 'isOwner'. Name override should be removed

PR #545, opened 2023-10-20
Feat(Gate): Renamed the field "isOwner" to "isOwnCompanyData"

cezaralexandremorais:
Renamed the field "isOwner" to "isOwnCompanyData" in OpenAPI

solves #510


bpdm-gate-api/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/api/model/IBaseBusinessPartnerGateDto.kt
```diff
@@ -28,6 +28,6 @@ interface IBaseBusinessPartnerGateDto : IBaseBusinessPartnerDto {
     @get:Schema(description = CommonDescription.externalId)
     val externalId: String
 
-    @get:Schema(name = "isOwner", description = "True if the sharing member declares itself as the owner of the business partner.")
+    @get:Schema(name = "isOwnCompanyData", description = "True if the sharing member declares itself as the owner of the business partner.")
```
nicoprow, line 31: better to just remove the whole name override. No need to have a name override there. Springdoc should just take the name of the field instead

nicoprow (changes requested):

fabiodmota (approved):

nicoprow (approved):

Issue #579, opened 2023-10-26
Refactor: Remove unnecessary imports


nicoprow:
We should remove unnecessary import directives from our code to enhance clarity.

Issue #578, opened 2023-10-26
Refactor: Remove unused symbols


nicoprow:
We have a lot of unused symbols in our code. We should remove those that really redundant to enhance code clarity.

<img width="496" alt="grafik" src="https://github.com/eclipse-tractusx/bpdm/assets/95762387/0cb33417-6426-437b-a29e-302d6a307ac7">

### Tasks

Remove unused symbols in 
- [x] BPDM Pool
- [x] BPDM Gate
- [x] BPDM Orchestrator
- [x] BPDM Cleaning Service Dummy

Issue #576, opened 2023-10-26
Refactor: Remove unnecessary not-null assertions


nicoprow:
In the BPDM applications we should remove all unnecessary not-null assertions to enhance the clarity of the coee.

Here is what an excerpt from a  quick analysis I made:

<img width="524" alt="grafik" src="https://github.com/eclipse-tractusx/bpdm/assets/95762387/1f5bbbef-401d-4e38-8a2b-eacd8116dc3d">

PR #602, opened 2023-11-03
Refactor(BPDM): Clean import and unused vars for code smelling

fabiodmota:
## Description
General code improvement in all Modules.
Removal of imports.
Removal of unused code and variables.
Logic improvement in separate ifs with equal conditions.
Removal of unnecessary assert not null.

Issues: https://github.com/eclipse-tractusx/bpdm/issues/578 https://github.com/eclipse-tractusx/bpdm/issues/579 https://github.com/eclipse-tractusx/bpdm/issues/576

## Pre-review checks

### DIFF
--- a/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/util/CreateSQLScriptToInsertRegions.kt
+++ b/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/util/CreateSQLScriptToInsertRegions.kt
@@ -68,7 +68,7 @@ class CreateSQLScriptToInsertRegions {
  * Second main conflicts with spring boot repackage. Uncomment to start generation.
  */
 /*
-fun main() {
+fun generate() {
 
     val regionsCsvStream = CreateSQLScriptToInsertRegions::class.java.getResourceAsStream("/regions/IP2LOCATION-ISO3166-2.CSV")
     if (regionsCsvStream != null) {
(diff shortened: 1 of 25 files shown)


nicoprow (changes requested):
Looks good!

The commit though has the wrong type. instead of starting with 'feat' should be 'refactor'

fabiodmota:
> Looks good!
> 
> The commit though has the wrong type. instead of starting with 'feat' should be 'refactor'

Thanks for the correction , I runned up with sonar and Trivy locally before cleaning

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/util/CreateSQLScriptToInsertRegions.kt
```diff
@@ -63,16 +63,4 @@ class CreateSQLScriptToInsertRegions {
 
 }
 
-/**
```
martinfkaeser, line 66: Although this whole utility program is not currently used, it was kept for recreating a SQL file if needed in the future.
The main() is part of the program but was commented out because Spring Boot doesn't like two main functions.
So I wouldn't just remove the whole function.

Maybe rename the function to `generate()` and comment it in with a remark that one can add a very simple main() function just calling generate().

fabiodmota: Done it , putted again with generate and leave the comment about the function on top, will request review again

martinfkaeser (approved):

nicoprow (changes requested):
There is one test of the Pool failing due to a rounding error it seems. Please have a look and correct the

fabiodmota:
> There is one test of the Pool failing due to a rounding error it seems. Please have a look and correct the

All passed and rebased on current main

nicoprow (approved):

Issue #431, opened 2023-09-01
Gate: Request Cleaning for BPN Update


nicoprow:
``Context: When business partner information in the BPDM Pool has been updated, Gates which have matching business partners need those business partners to be updated as well.``

The Gate should poll the BPN changelog of the Pool and check for updates on BPNs. If there are updates for BPNs which are referenced by business partners in the output stage, the Gate should request cleaning for this output data from the Orchestrator.

### Check whether business partner update needed
For a BPN update the Gate will request a cleaning to the orchestrator for business partners if he business partner has an associated BPN matching the updated BPN

### Request cleaning
If the business partner needs to be cleaned, the Gate should request a cleaning from the orchestrator. The request should be in mode NoChanges and the **output**  data needs to be put as the business partner data to clean.

### Tasks
- [x] Poll Pool BPN Changelog for new changes
- [x] Create cleaning task for business partners with Golden record updates (see #431)

Dependencies
#429

nicoprow:
In order to create the polling the Gate needs to remember the last time the changelog has been polled. This can be analogous to how the Pool does its sync with the Opensearch:

https://github.com/eclipse-tractusx/bpdm/blob/main/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/entity/SyncRecord.kt

https://github.com/eclipse-tractusx/bpdm/blob/main/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/component/opensearch/impl/service/OpenSearchSyncStarterService.kt

nicoprow:
This issue handles the second use case in the orchestration layer: https://github.com/eclipse-tractusx/bpdm/issues/455#issuecomment-1717272993

nicoprow:
For now checking the timestamp is not needed. This would mean that the Gate which was responsible for causing the Pool Changelog Entry will also start this new request.

PR #611, opened 2023-11-05
feat(gate): Request Cleaning for BPN Update

cezaralexandremorais:
Request Cleaning for BPN Update
solves #431

### DIFF
--- a/bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/service/BaseSyncRecordService.kt
+++ b/bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/service/BaseSyncRecordService.kt
@@ -92,14 +92,14 @@ abstract class BaseSyncRecordService<SYNC_TYPE : Enum<*>, SYNC_RECORD : BaseSync
     }
 
     @Transactional(isolation = Isolation.SERIALIZABLE)
-    open fun setSynchronizationSuccess(type: SYNC_TYPE): SYNC_RECORD {
+    open fun setSynchronizationSuccess(type: SYNC_TYPE, finishedAt: Instant? = null): SYNC_RECORD {
         val record = getOrCreateRecord(type)
         if (record.status != SyncStatus.RUNNING)
             throw BpdmSyncStateException("Synchronization of type ${record.type} can't switch from state ${record.status} to ${SyncStatus.SUCCESS}.")
 
         logger.debug { "Set sync of type ${record.type} to status ${SyncStatus.SUCCESS}" }
 
-        record.finishedAt = Instant.now().truncatedTo(ChronoUnit.MICROS)
+        record.finishedAt = finishedAt ?: Instant.now().truncatedTo(ChronoUnit.MICROS)
         record.progress = 1f
         record.status = SyncStatus.SUCCESS
         record.errorDetails = null
--- a/bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/config/ClientsConfig.kt
+++ b/bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/config/ClientsConfig.kt
@@ -19,6 +19,8 @@
 
 package org.eclipse.tractusx.bpdm.gate.config
 
+import org.eclipse.tractusx.bpdm.pool.api.client.PoolApiClient
+import org.eclipse.tractusx.bpdm.pool.api.client.PoolClientImpl
 import org.eclipse.tractusx.orchestrator.api.client.OrchestrationApiClient
 import org.eclipse.tractusx.orchestrator.api.client.OrchestrationApiClientImpl
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
@@ -37,7 +39,7 @@ import java.util.function.Consumer
 
 
 @Configuration
-class OrchestratorClientConfig {
+class ClientsConfig {
 
     // Orchestrator-Client without authentication
     @Bean
@@ -50,6 +52,22 @@ class OrchestratorClientConfig {
         val url = orchestratorConfigProperties.baseUrl
         return OrchestrationApiClientImpl { webClientBuilder(url).build() }
     }
+
+    @Bean
+    @ConditionalOnProperty(
+        value = ["bpdm.gate-security.pool-security-enabled"],
+        havingValue = "false",
+        matchIfMissing = true
+    )
+    fun poolClientNoAuth(poolConfigProperties: PoolConfigProperties): PoolApiClient {
+        val url = poolConfigProperties.baseUrl
+        return PoolClientImpl {
+            WebClient.builder()
+                .baseUrl(url)
+                .defaultHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE).build()
+        }
+    }
+
 
 
     @Bean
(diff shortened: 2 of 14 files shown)


bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -272,4 +280,51 @@ class BusinessPartnerService(
 
     }
 
+    @Scheduled(cron = "\${cleaningService.pollingCron:-}", zone = "UTC")
+    fun checkBpnUpdates() {
+
+        val poolPaginationRequest = PaginationRequest(0, 100)
```
nicoprow, line 286: the pagination size should best be configurable

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -272,4 +280,51 @@ class BusinessPartnerService(
...
+        val poolChangelogEntries = poolClient.changelogs.getChangelogEntries(changelogSearchRequest, poolPaginationRequest)
+        val poolUpdatedEntries = poolChangelogEntries.content.filter { it.changelogType.name == PoolChangeLogType.UPDATE.name }
+
+
+        val bpnA =
+            poolUpdatedEntries.filter { it.businessPartnerType == BusinessPartnerType.ADDRESS && it.changelogType == PoolChangeLogType.UPDATE }.map { it.bpn }
+        val bpnL = poolUpdatedEntries.filter { it.businessPartnerType == BusinessPartnerType.LEGAL_ENTITY && it.changelogType == PoolChangeLogType.UPDATE }
+            .map { it.bpn }
+        val bpnS =
+            poolUpdatedEntries.filter { it.businessPartnerType == BusinessPartnerType.SITE && it.changelogType == PoolChangeLogType.UPDATE }.map { it.bpn }
```
nicoprow, line 302: Didn't you already filter for ChangelogType UPDATE? Why do you check for the UPDATE type here again?

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -272,4 +280,51 @@ class BusinessPartnerService(
...
+                        partner.parentType!!
+                    ), taskId
+                )
+            }
+        sharingStateService.setPending(pendingRequests)
+
+        if (poolUpdatedEntries.isNotEmpty()) {
+            syncRecord.finishedAt = poolUpdatedEntries.last().timestamp
+            syncRecord.status = SyncStatus.SUCCESS
+            syncRecordService.setSyncRecord(syncRecord)
```
nicoprow, line 326: Ok, I assume you don't want to use the setSuccess method here because you want to overwrite the finishedAt timestamp. I think it would be cleaner and less error prone if you make the finishedAt-Overwrite an optional parameter in the setSuccess method and use that method here

nicoprow (changes requested):

bpdm-common/src/main/kotlin/org/eclipse/tractusx/bpdm/common/service/BaseSyncRecordService.kt
```diff
@@ -92,14 +92,14 @@ abstract class BaseSyncRecordService<SYNC_TYPE : Enum<*>, SYNC_RECORD : BaseSync
     }
 
     @Transactional(isolation = Isolation.SERIALIZABLE)
-    open fun setSynchronizationSuccess(type: SYNC_TYPE): SYNC_RECORD {
+    open fun setSynchronizationSuccess(type: SYNC_TYPE, finishedAt: Instant?): SYNC_RECORD {
```
nicoprow, line 95: This will lead to other overridden functions to fail. If you have a default null parameter, this is resolved

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/config/PoolConfigProperties.kt
```diff
@@ -25,4 +25,5 @@ import org.springframework.boot.context.properties.ConfigurationProperties
 @ConfigurationProperties(prefix = "bpdm.pool")
 data class PoolConfigProperties(
     val baseUrl: String = "http://localhost:8080/api/catena",
```
nicoprow, line 27: Base-URL is without api/catena

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/repository/generic/BusinessPartnerRepository.kt
```diff
@@ -35,4 +35,10 @@ interface BusinessPartnerRepository : JpaRepository<BusinessPartner, Long>, Crud
     fun findByStageAndExternalIdIn(stage: StageType, externalId: Collection<String>, pageable: Pageable): Page<BusinessPartner>
 
     fun findByStage(stage: StageType, pageable: Pageable): Page<BusinessPartner>
+    fun findByStageAndBpnLInOrBpnSInOrBpnAIn(
```
nicoprow, line 38: Does the logical 'or' take precedence over the 'and'? In any case I think its cleaner to write this in JPQL where you can set brackets how you need them.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -272,4 +281,51 @@ class BusinessPartnerService(
 
     }
 
+    @Scheduled(cron = "\${cleaningService.pollingCron:-}", zone = "UTC")
+    @Transactional
+    fun checkBpnUpdates() {
```
nicoprow, line 286: name of the method is a bit non-descriptive. "Check" not really conveys that it actually starts a golden record task if the check validates to true

nicoprow (changes requested):

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/config/PoolConfigProperties.kt
cezaralexandremorais: i used the same value  that already exists on application.properties , should i need to change it on application.properties too?

nicoprow: Yes please, the Gate was connected to the Pool before. Seems to be that was an outdated property.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/config/ClientsConfig.kt
```diff
@@ -51,6 +53,22 @@ class OrchestratorClientConfig {
         return OrchestrationApiClientImpl { webClientBuilder(url).build() }
     }
 
+    @Bean
+    @ConditionalOnProperty(
+        value = ["bpdm.gate-security.pool-security-enabled"],
+        havingValue = "false",
+        matchIfMissing = true
+    )
+    fun poolClientNoAuth(poolConfigProperties: PoolConfigProperties): PoolApiClient {
```
fabiodmota, line 62: We should add the auth profile for the pool since the pool on dev and int are using auth and by that the connection will fail

fabiodmota (changes requested):

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/config/ClientsConfig.kt
nicoprow: True, currently there is no pool client configuration in case the security is enabled. Should be introduced

nicoprow (changes requested):
Missing security configuration for an authorized pool client

cezaralexandremorais:
> Please squash the commits and then ready to merge

Done

Issue #573, opened 2023-10-26
Gate: Don't update Business Partners without change


nicoprow:
Currently, when business partner data for a given stage and external id should be upserted we should validate whether the business partner data really is different from the existing data in the database. Only when there is a change we should create change log entry. Likewise, only when business partner input is changed the Gate starts a golden record process.

### Tasks

- [x] Create validation logic comparing incoming business partner stage data with existing stage data
- [x] Only create changelog on changed data
- [x] Only create golden record tasks for changed data

fabiodmota:
When an input is made, it should be validated if it already exists in the persistence to avoid always triggering the clean flow among all components.

**Key Point:**
What happens when a `taskId` from the Orchestrator is returned to the Gate and for some reason, the Orchestrator loses context? We won't have our data cleaned, but the persistence of the Input will remain the same as initially and won't trigger new requests that come in identically.

**Recommendation:**
Implementing a validation mechanism to check the input's existence in the persistence layer before proceeding could mitigate such issues and ensure the system handles new and identical requests efficiently.

nicoprow:
> When an input is made, it should be validated if it already exists in the persistence to avoid always triggering the clean flow among all components.
> 
> **Key Point:** What happens when a `taskId` from the Orchestrator is returned to the Gate and for some reason, the Orchestrator loses context? We won't have our data cleaned, but the persistence of the Input will remain the same as initially and won't trigger new requests that come in identically.
> 
> **Recommendation:** Implementing a validation mechanism to check the input's existence in the persistence layer before proceeding could mitigate such issues and ensure the system handles new and identical requests efficiently.

The solution would be to the check the sharing state of the business partner whose data is being upserted. If the sharing state is in error a golden record task should be created regardless the input data is the same or not. If it is in Pending or in Success it should not be updated if the data is equal.

PR #615, opened 2023-11-08
Feat (Gate) - Validation for a Business Partner update

alexsilva-CGI:
## Description

This PR adds a validation for an update of a business Partner. If the update has the same data as the persisted Business Partner, it will not go through and as such no changelog and cleaning task will be created/started.

Also fixes a bug for updating states, classifications and identifiers. Bug was found in the copyAndSync method found in the CollectionsUtils.

Also the cron expression property anme found in the resolve of the cleaning task was not inline with the name found in the properties file. Naming was fixed.

Related issues: #573

## Pre-review checks


bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/AlternativePostalAddress.kt
```diff
@@ -72,4 +72,34 @@ class AlternativePostalAddress(
      */
     @Column(name = "alt_delivery_service_number")
     val deliveryServiceNumber: String?
-)
+) {
+    override fun equals(other: Any?): Boolean {
```
nicoprow, line 76: It's a neat idea but I'm not sure if we want to override the default equals method for this. Here, we define our own equals in a specific context. Like, it is not really completely equal since technical fields like timestamps and identifier could be still different. In other contexts we might want to treat the meaning of equality differently. I would propose to move to another service which contains the equality methods.

I'm open for discussion though, would like to see the input of the other reviewers

nicoprow (changes requested):

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/AlternativePostalAddress.kt
alexsilva-CGI: Firstly I was creating different methods which would compare the upserted object with the newly inserted object field by field, but thought that wasn't the right and cleanest way to do it. Asked for some tips from Fabio how could I do it a different way, and decided with this idea. Probably there is some other clean and easiest way to do such thing in Kotlin, but didn't found out. Any tips are welcome :)

martinfkaeser: If I'm not mistaken you're considering *all* properties for equality check.
There is no ID or timestamps because this is an Embeddable, not an Entity.
I would even make `AlternativePostalAddress` a data class like `GeographicCoordinate`, then we can get rid of manual `equals` and `hashCode` implementations.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/PhysicalPostalAddress.kt
```diff
@@ -113,4 +113,30 @@ class PhysicalPostalAddress(
      */
     @Column(name = "phy_door")
     var door: String?
-)
+) {
+
+    override fun equals(other: Any?): Boolean {
```
martinfkaeser, line 118: You can make this a data class and get rid of `equals`.
If not, take care that you also override `hashCode` together with `equals` in a compatible way!

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -110,7 +112,9 @@ class BusinessPartnerService(
         val externalIds = entityCandidates.map { it.externalId }
         assertInputStageExists(externalIds)
 
-        val resolutionResults = resolveCandidatesForStage(entityCandidates, StageType.Output)
+        val validatedEntities = filterUpdateCandidates(entityCandidates, StageType.Output)
```
martinfkaeser, line 115: I find it a little inconsistent that you are using one `filterUpdateCandidates` in `upsertBusinessPartnersInput` but the other in `upsertBusinessPartnersOutputFromCandidates`. Any reasons for that?

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -157,6 +161,27 @@ class BusinessPartnerService(
...
+     * Business Partner will start the process again.
+     */
+    private fun filterUpdateCandidates(entities: List<BusinessPartner>, stage: StageType): List<BusinessPartner> {
+        val externalIds = entities.map { it.externalId }
+        val persistedBusinessPartner = businessPartnerRepository.findByStageAndExternalIdIn(stage, externalIds)
+        val sharingStates = sharingStateRepository.findByExternalIdInAndBusinessPartnerType(externalIds, BusinessPartnerType.GENERIC)
+
+        return entities.filter { entity ->
+            val matchingBusinessPartner = persistedBusinessPartner.find { it.externalId == entity.externalId }
+            val matchingSharingStates = sharingStates.find { it.externalId == entity.externalId && it.sharingStateType == SharingStateType.Error }
```
martinfkaeser, line 179: You could simplify that to:
`val hasErrorSharingState = sharingStates.any { ... }`

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/generic/BusinessPartner.kt
```diff
@@ -92,4 +92,33 @@ class BusinessPartner(
     @Enumerated(EnumType.STRING)
     var parentType: BusinessPartnerType? = null
 
-) : BaseEntity()
+) : BaseEntity() {
+
+    override fun equals(other: Any?): Boolean {
```
martinfkaeser, line 97: For the `BusinessPartner` and `PostalAddress` I agree with Nico's opinion about overriding equals. Might make problems for JPA. Additionally it's dangerous to override equals but not hashCode.
Maybe better use a custom `hasRelevantChanges` function or something in the service.

martinfkaeser (changes requested):
About the issue how to compare BusinessPartners for *relevant changes* I would suggest this approach:
- For embeddables use `data classes`.
- For `BusinessPartner` and `PostalAddress` better implement `hasRelevantChanges()` as a service method.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/Street.kt
```diff
@@ -51,4 +51,23 @@ class Street(
     @Column
     val additionalNameSuffix: String? = null,
 
-    )
\ No newline at end of file
+    ) {
+
+    override fun equals(other: Any?): Boolean {
```
nicoprow, line 56: Was this method overlooked? I think the default equals for data classes can be used for the Street as well, correct?

nicoprow (changes requested):

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/Street.kt
martinfkaeser: And `Street` is not a data class yet.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
martinfkaeser: This question still seems to be open..

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -157,6 +161,53 @@ class BusinessPartnerService(
...
+     *
+     * Filters incoming Business Partner for changes against the same persisted record in the Database
+     * If the data is the same, the consequent Business Partner will not have a new changelog and the golden record
+     * process will not start. If Business Partner has the same data, but linked sharing state has an error state,
+     * Business Partner will start the process again.
+     */
+    private fun filterUpdateCandidates(entities: List<BusinessPartner>, stage: StageType): List<BusinessPartner> {
+        val externalIds = entities.map { it.externalId }
+        val persistedBusinessPartner = businessPartnerRepository.findByStageAndExternalIdIn(stage, externalIds)
+        val sharingStates = sharingStateRepository.findByExternalIdInAndBusinessPartnerType(externalIds, BusinessPartnerType.GENERIC)
```
martinfkaeser, line 175: You could change both lists to maps by adding `.associateBy({ it.external })`. This could make lookup simpler and more efficient.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -157,6 +161,53 @@ class BusinessPartnerService(
...
+                entity.stage != persistedBP.stage ||
+                entity.parentId != persistedBP.parentId ||
+                entity.parentType != persistedBP.parentType ||
+                entity.identifiers != persistedBP.identifiers ||
+                entity.states != persistedBP.states ||
+                entity.classifications != persistedBP.classifications ||
+                postalAddressCompare(entity.postalAddress, persistedBP.postalAddress)
+    }
+
+    private fun postalAddressCompare(entityPostalAddress: PostalAddress, persistedPostalAddress: PostalAddress): Boolean {
```
martinfkaeser, line 205: Maybe call it `postalAddressHasChanges`?

martinfkaeser (changes requested):

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
alexsilva-CGI: Yeah, basically in the upsertBusinessPartnersOutput is currently not being used, because the finishCleaningTask maps directly to a BusinessPartner and not to BusinessPartnerOutputRequest which is used in this method. This was done because otherwise two mappings we're needed. Probably some refactoring later will need to be done.

nicoprow (changes requested):
Looks good to me now! Just need to rebase to current main and squash the existing commits.

martinfkaeser (approved):

Issue #588, opened 2023-10-31
Orchestrator: Service Logic for Timeout


nicoprow:
When a task reaches its timeout limit it should set into an error state with the errorcode 'Timeout'

martinfkaeser:
In the refinement session we further discussed the details.

Two different kinds of timeout should be checked:
- A task remaining in state "Pending" for a specified time should go into state "Error" with error type "Timeout": *TaskPendingTimeout*
- A specified time after reaching its final state ("Success" or "Error") the task should be completely removed from the Orchestrator: *TaskRetentationTime*

martinfkaeser:
During implementation this problem came up:
In our current API we return just *one timeout* when reserving *several tasks* for cleaning.
But each task can have a different timeout, because it is calculated as **task.creationTime** + **taskPendingTimeout**.

So I would suggestion moving the timeout down to the task data.
Now I realized in ReDoc the API has already been changed exactly like that:
https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/eclipse-tractusx/bpdm/docs/api/generic-input/docs/api/orchestrator-proposal.yaml#tag/Cleaning-Steps/paths/~1cleaning-tasks~1reservation/post
So I'll update the implementation.

nicoprow:
Ok, seems to be we did not ahere to the OpenAPI spec completely in that regard. Nevertheless we should e careful about introducing breaking changes into the BPDM applications. If the feature will need to be a breaking change the time when we can merge it may increase. A minor version update could be simpler: Just leave the global timeout and add the timeout field for each indiviudal task. 

We should plan ahead when we want to introduce the next breaking change to see how much we need to consider this aspect for the feature

martinfkaeser:
Okay, I'll mark the global *timeout* field as deprecated.

I'm wondering if in the reservation result a timeout even makes sense or if we should remove it completely?
After all it is not only the timeout for posting the result but for completely processing the task.
If a cleaning service posts the cleaning results e.g. 10 secs before the announced timeout this will probably not be enough because the Pool still needs to process the task in the remaining time. 
The problem would be exacerbated if there are multiple cleaning steps.

What do you think, @nicoprow ?

PR #617, opened 2023-11-09
feat(orchestrator): Service logic for timeout

martinfkaeser:
In the Orchestrator a cron-job checks if any of the two timeouts have been reached for any of the tasks:
- A task remaining in state "Pending" for a specified time (*taskPendingTimeout*) after its creation should go into state "Error" with error type "Timeout".
- A specified time (*taskRetentionTimeout*) after its creation a task should be completely removed from the Orchestrator.

**TaskStepReservationResponse.timeout** has been marked as deprecated as this field doesn't make much sense for a cleaning service.
We might add a task specific timeout in **TaskStepReservationEntryDto** but this is still up for debate.

Integration tests for the timeout cases have been added.

Solves #588

### DIFF
--- a/bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/GoldenRecordTaskService.kt
+++ b/bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/GoldenRecordTaskService.kt
@@ -19,11 +19,14 @@
 
 package org.eclipse.tractusx.bpdm.orchestrator.service
 
+import mu.KotlinLogging
+import org.eclipse.tractusx.bpdm.orchestrator.config.TaskConfigProperties
 import org.eclipse.tractusx.bpdm.orchestrator.exception.BpdmEmptyResultException
 import org.eclipse.tractusx.bpdm.orchestrator.exception.BpdmTaskNotFoundException
 import org.eclipse.tractusx.bpdm.orchestrator.model.GoldenRecordTask
 import org.eclipse.tractusx.bpdm.orchestrator.model.TaskProcessingState
 import org.eclipse.tractusx.orchestrator.api.model.*
+import org.springframework.scheduling.annotation.Scheduled
 import org.springframework.stereotype.Service
 import java.time.Instant
 import java.util.*
@@ -31,8 +34,11 @@ import java.util.*
 @Service
 class GoldenRecordTaskService(
     val taskStorage: GoldenRecordTaskStorage,
-    val goldenRecordTaskStateMachine: GoldenRecordTaskStateMachine
+    val goldenRecordTaskStateMachine: GoldenRecordTaskStateMachine,
+    val taskConfigProperties: TaskConfigProperties
 ) {
+
+    private val logger = KotlinLogging.logger { }
 
     @Synchronized
     fun createTasks(createRequest: TaskCreateRequest): TaskCreateResponse {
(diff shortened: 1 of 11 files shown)


bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/GoldenRecordTaskService.kt
```diff
@@ -95,6 +98,24 @@ class GoldenRecordTaskService(
...
 
+    @Scheduled(cron = "\${bpdm.task.timeoutCheckCron}")
+    @Synchronized
+    fun checkForTimeouts() {
+        logger.debug { "Checking for timeouts" }
+
+        taskStorage.getTasksCreatedBefore(Instant.now().minus(taskConfigProperties.taskRetentionTimeout))
+            .forEach {
+                logger.info { "Removing task ${it.taskId} after retention timeout" }
+                taskStorage.removeTask(it.taskId)
```
nicoprow, line 109: I would propose to not start the retention time period with the creation time but with the instant the task has finished processing. This could be a success or error. With the pending timeout in place a task is guaranteed to end in such a final state, so the retention period could start from there

nicoprow (changes requested):

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/GoldenRecordTaskService.kt
martinfkaeser: This was my initial idea too and that is what we discussed in the refinement.

But on starting implementation I realized this poses a problem to the current API: 
Upon creating a new task the caller gets back a timeout ("The timestamp until the task is removed from the Orchestrator"). 
But this is unknown upon creation if we define the retention time only starts with the task reaching its final state.
This is what I added as a comment: #https://github.com/eclipse-tractusx/bpdm/issues/588#issuecomment-1801550090

So in my opinion we have two options:
1) We deprecate and later remove the `timeout` property in `TaskProcessingStateDto`. I can't see that it's particularly useful for the caller anyways. We already deprecated `timeout` in `TaskStepReservationResponse`.
2) We define the retention period as starting with the task creation, as implemented. In practice this doesn't make a big difference if the TaskRetentionTimeout is big compared to TaskPendingTimeout.
Maybe 1) is the cleaner way.

nicoprow: Yea I would go with the first choice I think. I would expect the timout only to be the pending timeout anyway, so we could avoid sending it back to the task creator as its of limited use for him. 

Or there is also a third option:

The overall time until a task is deleted is the pending time plus retention time. So even if a task is resolved sooner than the pending time, a task will always be available the total of pending + retention time

martinfkaeser: Yes, but with the third option we would get the same result as implemented now, just then deletionTimeout = pendingTimeout + rententionTimeout.

I'll go with option 1).

martinfkaeser: Implemented

We should take into considering the two deprecated `timeout` properties regarding the Standard.

nicoprow (approved):

PR #622, opened 2023-11-16
Feat(Charts): Remove Opensearch

nicoprow:
## Description

This pull requests removes the Opensearch dependency from the BPDM Umbrella and BPDM Pool Charts. Opensearch is not used anymore in the BPDM applications, so this is not regarded as a breaking change. 

## Pre-review checks


nicoprow:
Currently deployment of chart fails as it still references app version 4.1.0 which has the BPDM Pool that relies on an existing Opensearch deployment. 

Need to wait for the release of 4.1.1 which will not include Opensearch anymore and reference that one in the chart.

Issue #572, opened 2023-10-26
Refactor: BPDM applications properties structure


nicoprow:
Currently, the BPDM applications have a bit of a messy property naming/pathing. We should have a clear and easy naming conventions for our properties, so they can be easily understood and grouped.

My proposal:

- [x] All properties we have defined should prefix with "bpdm"
- [x] All properties which configure the security of the own API should prefix with "bpdm.security"
- [x] The permissions that are used for the own API authorization should prefix with "bpdm.security.permission"
- [x] The properties configuring how the BPDM clients for accessing other BPDM applications should prefix with "bpdm.client._CLIENT_NAME_", so for orchestrator client configuration the prefix should "bpdm.client.orchestrator"
- [x] The application should have default properties with security disabled and clients security disabled
- [x] There is a Profile auth which activates and configures the application security
- [x] There are profiles _CLIENTNAME_-auth which activates security configuration for its respective client (so orchestrator-auth for orchestrator security)

Issue #567, opened 2023-10-25
Properties: Convert application properties to yaml


nicoprow:
In our supported deployment context we use Helm and Kubernetes which are very yaml focused. In fact, we also have the possibiltity to override application properties through our Helm Charts via yaml format. For this reason I propose to convert the existing application.properties files to the yaml format. This helps to better integrate with the Kubernetes context.

PR #623, opened 2023-11-16
Refactor (bpdm) - Conversion of properties files to YML 

alexsilva-CGI:
In this PR, application properties we're converted to YML and some properties name path's we're change accordingly.

Solve #572 and #567

## Pre-review checks


nicoprow (approved):
Looks good to me.

Please not that this pull request contributes to #572 but it does not solve it completely. We would need follow-up pull requests for renaming the paths for permissions and introduce separate auth yaml files for the clients.

nicoprow (changes requested):
Please squash commits and rebase, then the pull request is ready to merged

Issue #600, opened 2023-11-03
Security: Adapt Security to New Spring Version 


fabiodmota:
Security configuration update.
Since the update of the new version of Spring, the configuration class is marked as deprecated, it is necessary to see what the new methods are and ways to maintain the same settings with the new configuration.
This update needs to be done for all the modules that have the class.

![image](https://github.com/eclipse-tractusx/bpdm/assets/107047355/6b866186-1b87-4835-9ed9-461f5ddf66cf)

PR #628, opened 2023-11-16
fix: Adapt Spring Security configuration to new Spring version

martinfkaeser:
Adapt `BpdmSecurityConfigurerAdapter` configuration for each module to the new Spring version, there some of the old methods where deprecated.

Solves #600

### DIFF
--- a/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/config/BpdmSecurityConfigurerAdapterImpl.kt
+++ b/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/config/BpdmSecurityConfigurerAdapterImpl.kt
@@ -26,34 +26,30 @@ import org.springframework.context.annotation.Configuration
 import org.springframework.http.HttpMethod
 import org.springframework.security.config.annotation.web.builders.HttpSecurity
 import org.springframework.security.config.http.SessionCreationPolicy
-import org.springframework.security.web.util.matcher.AntPathRequestMatcher
+import org.springframework.security.web.util.matcher.AntPathRequestMatcher.antMatcher
 
 @Configuration
 class BpdmSecurityConfigurerAdapterImpl(
     val securityConfigProperties: SecurityConfigProperties
 ) : BpdmSecurityConfigurerAdapter {
 
-    companion object {
-        const val ALL_API_PATHS = "/api/**"
-    }
-
     override fun configure(http: HttpSecurity) {
-        http.csrf().disable()
-            .cors()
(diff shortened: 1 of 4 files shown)


bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/config/BpdmSecurityConfigurerAdapterImpl.kt
```diff
@@ -38,22 +38,25 @@ class BpdmSecurityConfigurerAdapterImpl(
...
+            it.requestMatchers(AntPathRequestMatcher(ALL_API_PATHS, HttpMethod.OPTIONS.name())).permitAll()
+            it.requestMatchers(AntPathRequestMatcher("/")).permitAll() // forwards to swagger
+            it.requestMatchers(AntPathRequestMatcher("/docs/api-docs/**")).permitAll()
+            it.requestMatchers(AntPathRequestMatcher("/ui/swagger-ui/**")).permitAll()
+            it.requestMatchers(AntPathRequestMatcher("/actuator/health/**")).permitAll()
+            it.requestMatchers(AntPathRequestMatcher(ALL_API_PATHS, HttpMethod.GET.name())).authenticated()
+            it.requestMatchers(AntPathRequestMatcher("/api/catena/**/search", HttpMethod.POST.name())).authenticated()
+            it.requestMatchers(AntPathRequestMatcher("/api/catena/**/filter", HttpMethod.POST.name())).authenticated()
+            it.requestMatchers(AntPathRequestMatcher("/error")).permitAll()
+            it.requestMatchers(AntPathRequestMatcher(ALL_API_PATHS)).hasRole("add_company_data")
```
nicoprow, line 54: I think in the scope of this pull request we can also align the Pool's authorization behaviour. I would expect the same as the Gate here, since we now configure authorization of each endpoint individually over the PreAuthorize annotation. Everything that is not covered by a PreAuthorize annotation would then be at least authenticated.

nicoprow (changes requested):
I propose to update the Pool's authorization configuration to be the same as the Gate's, as a detailed authorization through requestMatchers is no longer needed in the Pool.

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/config/BpdmSecurityConfigurerAdapterImpl.kt
martinfkaeser: Done.
Also I changed AntPathRequestMatcher to the antMatcher() shortcut.

fabiodmota (approved):

nicoprow (approved):

Issue #570, opened 2023-10-26
Refactoring: Business Partner DTO Locations


nicoprow:
We should refactor the DTO models in the BPDM modules. DTO classes are currently residing in various modules and there is no clear rule yet. Also not all Business Partner DTO classes implement from the Base DTO interfaces making sure that they are aligned in regards to field names and OpenAPI descriptions.

1. Business Partner DTOs should inherit from Base DTO interfaces
2. The Business Partner DTOs should be located in their own BPDM module (not in the common)

- [x] Reorganize Business Partner DTOs
- [x] Reorganize Legal Entity DTOs
- [x] Reorganize Site DTOs
- [x] Reorganize Address DTOs

PR #635, opened 2023-11-22
Refactoring: Generic Business Partner DTO Locations

martinfkaeser:
The DTOs related to the Generic Business Partner were refactored, making sure all its data classes are located in the appropriate API modules (gate-api and orchestrator-api). Common interfaces are used whenever possible.
This meant splitting up some DTOs for the different modules which were formerly in the common module.

This is the first step towards #570


alexsilva-CGI (approved):
Seems fine by me as well.

Issue #566, opened 2023-10-25
Logging: Set Logging Pattern in application.properties


nicoprow:
Currently, the logging patterns of the different BPDM modules is not uniform. Each module should have a standard pattern that can be overwritten over the application.properties. 

The standard pattern be can be taken from the logback configuration file in the BPDM pool module.

### Tasks
- [ ] For each module: Configure logback over the application.properties with the standard pattern
- [ ] Delete the separate logback configuration file

Issue #565, opened 2023-10-25
Logging: Add DEBUG Logging Message to Services


nicoprow:
Currently we have some gaps in our logging when it comes to normal logic execution. Some services log a lot while others are completely silent. I propose we should add a DEBUG Level log message at the beginning of each public service class method. The log message should contain what kind of operation is starting to be executed and the most important parameters of that operation.

PR #637, opened 2023-11-23
Refactor (bpdm) - DEBUG Logging Messages and Logging Patterns

alexsilva-CGI:
In this PR, DEBUG Logging Messages to Services which didn't have them before we're added. Also, logging Patterns in the application.yml we're added.

Solve #566 and #565

Should be merged only after https://github.com/eclipse-tractusx/bpdm/pull/623 PR

## Pre-review checks


bpdm-cleaning-service-dummy/src/main/resources/application.yml
```diff
@@ -29,27 +29,34 @@ bpdm:
...
     cleaningService:
-        pollingCron: '-'
+      pollingCron: '-'
+
+# Logging Configuration
+logging:
+  unknown-user: Anonymous
+  pattern: "%clr(%d{yyyy-MM-dd HH:mm:ss.SSS}){faint} %clr(%5p) %clr([%-20.20X{user:-System}]){blue} %clr([%-36.36X{request:-No Request}]){magenta} %clr([%-15.15t]){faint} %clr(%-40.40logger{39}){cyan} %clr(:){faint} %m%n%wEx"
+  level:
+    root: DEBUG
```
nicoprow, line 47: But DEBUG on root level would mean all messages are DEBUG on default, right? How about making root INFO and make the BPDM package name DEBUG?

alexsilva-CGI: Did the change. Would the logging level's be something like that?

bpdm-bridge-dummy/src/main/resources/application-auth.yml
```diff
@@ -24,6 +24,17 @@ bpdm:
     refresh-url: ${bpdm.security.token-url}
     token-url: ${bpdm.security.auth-server-url}/realms/${bpdm.security.realm}/protocol/openid-connect/token
 
+# Logging Configuration
+logging:
+  pattern: "%clr(%d{yyyy-MM-dd HH:mm:ss.SSS}){faint} %clr(%5p) %clr([%-15.15t]){faint} %clr(%-40.40logger{39}){cyan} %clr(:){faint} %m%n%wEx"
```
nicoprow, line 29: As far as I have tested it, the pattern is not being taking into account. I guess it's because it should be the pattern for the console output. That is: 
logging:
   pattern:
     console:

According to my tests putting the pattern there will lead to it being recognized. I also discovered that the %clr command is not working on this logger. Instead we should use direct color commands. I'm open to any color pattern. Here is a suggestion:

%d{yyyy-MM-dd HH:mm:ss.SSS} %highlight(%-5level) %green([%-10.10X{user:-System}]) %magenta([%-10.10X{request:-No Request}]) %yellow([%-15.15t]) %cyan(%-40.40logger{39}) : %m%n%wEx

nicoprow (changes requested):
Logging Level works as expected but logging pattern is not taken into account.

Otherwise needs to have a squash for the commits.

bpdm-bridge-dummy/src/main/resources/application-auth.yml
alexsilva-CGI: That color pattern looks good, at least for me. Regarding the Auth properties, should they have something different of the default properties pattern?

nicoprow: Pattern is being recognized by Spring Boot now, so that's good. I just noticed the auth profile also has its own pattern. Please remove those changes, the auth profile does not need its own debug behaviour, it can just reuse the default properties

nicoprow (changes requested):
Pattern looks good now.

But please remove the logging overwrite in the application-auth. We don't need a separate logging behaviour for the auth profile. It can just take the default properties.

nicoprow (approved):

nicoprow (changes requested):
@alexsilva-CGI please fix the conflicts that resulted from the previous merges to main

nicoprow (approved):

Issue #657, opened 2023-11-29
Pool: LegalAddress can be Main Site Address


nicoprow:
Currently, it is not possible in the Pool for a legal address to be also a main site address at the same time. However, according to the BPDM Gate and the golden record process this should be possible. We need to adapt the Pool so a legal address can be a site main address at the same time.

alexsilva-CGI:
I have some doubts regarding this @nicoprow . Currently if an Legal Entity is created and a Site, with the same Address information, two different rows are created on the DB with different BPN's and associations. It seems that it is possible the creation of an Address with both Legal Entity and Site as there are no constraints limiting that.

How would we go, logic wise, to adapt the pool to know if the associated legal address can also be an site main address? Probably by first checking if there similar addresses (Same values) persisted, or is there a field that can be used for this association?

nicoprow:
> I have some doubts regarding this @nicoprow . Currently if an Legal Entity is created and a Site, with the same Address information, two different rows are created on the DB with different BPN's and associations. It seems that it is possible the creation of an Address with both Legal Entity and Site as there are no constraints limiting that.
> 
> How would we go, logic wise, to adapt the pool to know if the associated legal address can also be an site main address? Probably by first checking if there similar addresses (Same values) persisted, or is there a field that can be used for this association?

The Pool doesn't need to infer by itself whether an address is a legal address and/or a site main address. It should be possible to read that from the orchestrator business partner when processing a golden record ticket. There, the address type clearly states whether the address is a legal address, site main address or both.

When it comes to the previous way on how we created business partners in the Pool through the POST and PUT endpoints, I don't think we need to support them anymore. The logic for them can be as is. This is just about the new integration logic with the orchestrator.

PR #674, opened 2023-12-06
Fix (Pool) - Added way to LegalAddress to be Main Site Address in Pool

alexsilva-CGI:
## Description

Currently, it is not possible in the Pool for a legal address to be also a main site address at the same time. This PR adjust the way a Address is persisted if both Legal Entity and Site share it.

Related issues: #657 

## Pre-review checks

### DIFF
--- a/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/TaskStepBuildService.kt
+++ b/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/TaskStepBuildService.kt
@@ -20,6 +20,7 @@
 package org.eclipse.tractusx.bpdm.pool.service
 
 import jakarta.transaction.Transactional
+import org.eclipse.tractusx.bpdm.common.dto.AddressType
 import org.eclipse.tractusx.bpdm.common.dto.BusinessPartnerType
 import org.eclipse.tractusx.bpdm.pool.api.model.ChangelogType
 import org.eclipse.tractusx.bpdm.pool.dto.AddressMetadataDto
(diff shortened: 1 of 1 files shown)


alexsilva-CGI:
@nicoprow Tested with the round trip, and it seems that its now associating both site and legalEntity to an address when the LegalAndSiteMainAddress is present on the AddressType. 

I do have a question. Should also this logic apply when an update is done? so for example when type changes from SiteMainAddress to LegalAndSiteMainAddress, an association from both Site and Legal Entity to a single Address is done, with the deletion of the old persisted addresses. Thanks

nicoprow:
> @nicoprow Tested with the round trip, and it seems that its now associating both site and legalEntity to an address when the LegalAndSiteMainAddress is present on the AddressType.
> 
> I do have a question. Should also this logic apply when an update is done? so for example when type changes from SiteMainAddress to LegalAndSiteMainAddress, an association from both Site and Legal Entity to a single Address is done, with the deletion of the old persisted addresses. Thanks

For the update I would follow the same principle we have done before: The parent relationship between L/S/A business partners doesn't change

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/TaskStepBuildService.kt
```diff
@@ -296,10 +297,19 @@ class TaskStepBuildService(
             val bpnS = bpnIssuingService.issueSiteBpns(1).single()
             val createSite = BusinessPartnerBuildService.createSite(siteDto, bpnS, legalEntity)
             taskEntryBpnMapping.addMapping(bpnSReference, bpnS)
-            val address = createLogisticAddress(mainAddress, taskEntryBpnMapping)
-            createSite.mainAddress = address
-            address.site = createSite
-            createSite
+
+            if (taskEntry.generic.postalAddress.addressType == AddressType.LegalAndSiteMainAddress) { //Association of Site Main Address to Legal Address
```
nicoprow, line 301: There is some duplicated code here. Why not somthing like:

```
val siteMainAddress = 
            if(genericBusinessPartner.postalAddress.addressType == AddressType.LegalAndSiteMainAddress) 
                legalEntity.legalAddress 
            else  
                createLogisticAddress(mainAddress, taskEntryBpnMapping)
        
  createSite.mainAddress = siteMainAddress
  siteMainAddress.site = createSite
  createSite
```

nicoprow (approved):

nicoprow (changes requested):
See the comment

dilipdhankecha2530:
@nicoprow @alexsilva-CGI 
We also need to think for the new endpoints which will do the same. In the current development there is no endpoint which we can use to create LegalEntity address as a SiteAddress.

PR #675, opened 2023-12-06
Refactoring: Reorganize site & address DTOs between modules

martinfkaeser:
The DTOs related to sites and logistic addresses were refactored, making sure all its data classes are located in the appropriate API modules (gate-api, pool-api and orchestrator-api). Common interfaces are used whenever possible.
This meant splitting up and adjusting some DTOs for the different modules which were formerly in the common module.

Also some formerly common DTOs that are only used by the Pool module were relocated, e.g. some metadata DTOs.

This is another step towards https://github.com/eclipse-tractusx/bpdm/issues/570


nicoprow (changes requested):
And new conflict again ^^

nicoprow (approved):

PR #678, opened 2023-12-08
Feat/orchestrator pool validate improvement

rainer-exxcellent:
## Description
Improve the validation for the golden record process so that it:

- validats duplicates over all addresses and all legal entities
- if the BPN request identifier is known to the pool, the task has to be considered as update

### DIFF
--- a/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/RequestValidationService.kt
+++ b/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/RequestValidationService.kt
@@ -29,6 +29,9 @@ import org.eclipse.tractusx.bpdm.pool.api.model.request.*
 import org.eclipse.tractusx.bpdm.pool.api.model.response.*
 import org.eclipse.tractusx.bpdm.pool.dto.LegalEntityMetadataDto
 import org.eclipse.tractusx.bpdm.pool.repository.*
+import org.eclipse.tractusx.bpdm.pool.service.TaskStepBuildService.CleaningError.LEGAL_ENTITY_IS_NULL
+import org.eclipse.tractusx.orchestrator.api.model.BusinessPartnerFullDto
+import org.eclipse.tractusx.orchestrator.api.model.LogisticAddressDto
 import org.eclipse.tractusx.orchestrator.api.model.TaskStepReservationEntryDto
 import org.springframework.stereotype.Service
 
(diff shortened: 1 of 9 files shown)


nicoprow (approved):

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/RequestValidationService.kt
```diff
@@ -43,109 +45,201 @@ class RequestValidationService(
...
+
+        return mergeMapsWithCollectionInValue(siteErrorsByRequest, mainAddressEmptyErrorsByRequest)
     }
 
-    fun validateSitesToCreateFromOrchestrator(taskReservationRequests: List<TaskStepReservationEntryDto>): Map<RequestWithKey, Collection<ErrorInfo<SiteCreateError>>> {
+    private fun validateAddressesFromOrchestrator(taskEntries: List<TaskStepReservationEntryDto>,  requestMappings: TaskEntryBpnMapping): Map<RequestWithKey, Collection<ErrorInfo<ErrorCode>>> {
 
-        // parents have to be checked in the legal entity validation
-        val (requestsWithMainAddressNull, requestsWithMainAddressNotNull) = taskReservationRequests.partition { it.businessPartner.site?.mainAddress == null }
+        val taskWithLegalAddress = taskEntries
```
martinfkaeser, line 134: Why this filter condition? 
Is it only about legal addresses, as the variable name suggests?
What about pure additional addresses, not legal or main addresses?
I find this address validation method and its sub-methods confusing.

martinfkaeser (changes requested):
Apart from some minor remarks my main point is about the `validateAddressesFromOrchestrator` method and its sub-methods. I find the logic there quite confusing. Could you maybe have a look at it again and add some comments about the different cases?
And could you make sure you set the reformat option before commit because there are some weird spaces and indentations?

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/RequestValidationService.kt
rainer-exxcellent: This filter is a precondition for the following methods. Only addresses that have a LegalAddress and that have a MainAdress if a site is present may be considered.
As mentioned in the description , the check for duplicates has to be done over all inserted/updated addresses. So we have to collect first alle addresses.

martinfkaeser (approved):
This validation logic is inherently rather complex, but I think now we have reached, relative to that, quite an elegant solution! 👍

Issue #673, opened 2023-12-06
Pool: Street Name Additional Fields


nicoprow:
The street informationen in the Pool's address business partners V5 (#631 ) should now be extended by name prefix and suffix fields as they are already in the generic business partner model:

```mermaid
classDiagram

class StreetProperties{
        Street Name
        Street Name Prefix
        Street Name Additional Prefix
        Street Name Suffix
        Street House Number
        Street Direction
        Street Milestone
    }

```

PR #685, opened 2023-12-18
feat(pool): Added Street Name Additional Fields information

cezaralexandremorais:
Added Street Name Additional Fields information

Solves #673


nicoprow (changes requested):
Data Model looks good.

Testing the values is missing. I would expect the new street names added to the test values.

nicoprow (approved):

Issue #625, opened 2023-11-16
Logging: Add INFO Logging Message to Services


martinfkaeser:
We should find a common approach what events a service should log on level INFO.
I propose that any high level data change should be logged on INFO, e.g. "Legal entity BPNLxxx was created" or "Logistic address BPNAxxx was updated".

- [ ] Pool
- [ ] Gate
- [ ] Orchestrator

PR #720, opened 2024-01-25
feat: added Logging info to services

cezaralexandremorais:
added Logging info to services
solves #625


bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/AddressPersistenceService.kt
```diff
@@ -70,10 +72,12 @@ class AddressPersistenceService(
                 ?.let { existingAddress ->
                     updateAddress(existingAddress, address, legalEntityRecord, siteRecord)
                     gateAddressRepository.save(existingAddress)
+                    logger.info { "Address ${existingAddress.bpn} was updated" }
                     saveChangelog(address.externalId, ChangelogType.UPDATE, dataType)
                 }
                 ?: run {
                     gateAddressRepository.save(fullAddress)
+                    logger.info { "Address ${fullAddress.bpn} was updated" }
```
nicoprow, line 80: I guess should be logging "... was created" instead of "... was updated"

nicoprow (changes requested):
Small logging mistake otherwise looks good

nicoprow (approved):

PR #744, opened 2024-02-07
doc: Enhance documentation BPDM including Arc42 and ADRs regarding Value Added Services and EDC related topics for release 2403

gerlitmcoding:
## Description

This PR enhances documentation to the BPDM Solution. 
Included Topics are:
* The Trigger Configuration for start the golden record cleaning process
* The integration of value-added-services
* Documentation about EDC Asset Structuring
* (Maybe thoughts/solutions about FrameworkCredentials in context of BPDM)

## Pre-review checks


Sebastian-Wurm:
Good start. The requirement seems to be to have general control over the Sharing / Golden Record process. If this is the case, then we should already provide for this a future UI component in our architecture diagrams. In my opinion, it is not enough to simply describe a new call with an endpoint in a sequence diagram. We should rather describe in an ADR with reference to the corresponding business processes and requirements that such a UI component is required. It seems that the business processes are still to be defined, which means this change in the Arc42 seems to anticipate the requirements and business processes.

gerlitmcoding:
> Good start. The requirement seems to be to have general control over the Sharing / Golden Record process. If this is the case, then we should already provide for this a future UI component in our architecture diagrams. In my opinion, it is not enough to simply describe a new call with an endpoint in a sequence diagram. We should rather describe in an ADR with reference to the corresponding business processes and requirements that such a UI component is required. It seems that the business processes are still to be defined, which means this change in the Arc42 seems to anticipate the requirements and business processes.

In General I am with you. Since the corresonding business processes and requirements are not fully defined yet, I would see the mentioned adr for this in the next release. Fine for you?

Issue #627, opened 2023-11-16
Exception Handler: Handle Exceptions from malformed JSON requests


martinfkaeser:
We need a exception handler handling exceptions coming from the deserializing of JSON request payloads. We want to give the information why the JSON request could not be deserialized based on the information the JSONDeserializer gives us.

PR #753, opened 2024-02-14
Feat/exception handler

cezaralexandremorais:
added info about malformed JSon exception

solves #627


nicoprow (changes requested):
The error message still did not show up for me. Would need to add the information to the ResponseEntity or create a custom ResponseEntity

nicoprow (approved):
I guess I didn't correctly execute the changes. It works now for me perfectly

PR #777, opened 2024-02-23
Charts: Update to 4.0.1

### LINKED ISSUE #767: TRG 5.04 CPU / MEM resource requests and limits and are properly set
For https://eclipse-tractusx.github.io/docs/release/trg-5/trg-5-04/

CPU seems totally fine but Memory should be equal because of:

> Memory
Verify that memory request and limit are set equal. Memory is not a 'stretchable' resource and kubernetes will delete a pod if it starts using more memory than expected. Setting the memory limit to a higher value allows it to have a hidden memory profile which might work as long as the node has enough memory and will destabilize when the node does not have enough memory anymore.

found in:
https://github.com/eclipse-tractusx/bpdm/blob/eb05a1f02b11f6513d3545e6c29131efd4c4c70c/charts/bpdm/charts/bpdm-orchestrator/values.yaml#L66-L72

https://github.com/eclipse-tractusx/bpdm/blob/eb05a1f02b11f6513d3545e6c29131efd4c4c70c/charts/bpdm/charts/bpdm-bridge-dummy/values.yaml#L64-L70

https://github.com/eclipse-tractusx/bpdm/blob/eb05a1f02b11f6513d3545e6c29131efd4c4c70c/charts/bpdm/charts/bpdm-cleaning-service-dummy/values.yaml#L60-L66

nicoprow:
## Description

This pull request aligns the BPDM Chart's memory request and memory limit by setting both to 512Mi on default values.

Additionally, I removed some mentions of Opensearch configuration from the chart's READMEs.

Fixes #767 

## Pre-review checks

### DIFF
--- a/charts/bpdm/CHANGELOG.md
+++ b/charts/bpdm/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 The format is based on Keep a Changelog (https://keepachangelog.com/en/1.0.0/),
 
+## [4.0.1] - 2024-02-23
+
+### Changed
+
+- update BPDM Pool Chart to version 6.0.1
+- update BPDM Gate Chart to version 5.0.1
+- update BPDM Orchestrator Chart to version 2.0.1
+- update BPDM Cleaning Service Dummy Chart to version 2.0.1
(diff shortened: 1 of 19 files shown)


FaGru3n (approved):
LGTM

Issue #783, opened 2024-02-26
BPDM Bridge 500 error on incomplete alternative addresses


nicoprow:
Expected:
When sharing member records have not enough information in the alternative address so that they can't be written into the Pool, the Bridge should fail the import of that record and write the reason in the sharing state. Also, other records should be imported regardless.

Actual:

When trying to import such an incomplete record into the Pool the Bridge fails with a 500 and stops importing completely.

Example:

Alternative Address data given but no city.

PR #790, opened 2024-02-26
Bridge: Handle 500 Error 

fabiodmota:
## Description
Handle error on BPDM Bridge when sync from Gate To Pool
Exception is thrown when mapping logistic address

Fixes: https://github.com/eclipse-tractusx/bpdm/issues/783

## Pre-review checks

### DIFF
--- a/bpdm-bridge-dummy/src/main/kotlin/com/catenax/bpdm/bridge/dummy/service/PoolUpdateService.kt
+++ b/bpdm-bridge-dummy/src/main/kotlin/com/catenax/bpdm/bridge/dummy/service/PoolUpdateService.kt
@@ -22,6 +22,7 @@ package com.catenax.bpdm.bridge.dummy.service
 import com.catenax.bpdm.bridge.dummy.dto.*
 import mu.KotlinLogging
 import org.eclipse.tractusx.bpdm.common.dto.BusinessPartnerType
+import org.eclipse.tractusx.bpdm.common.exception.BpdmNullMappingException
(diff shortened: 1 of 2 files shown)


nicoprow (changes requested):
During my testing I still encounter this error because it happens when Legal Entities are created. This only fixes legal entities that are updated, but if someone creates a new legal entity the error would still be there

bpdm-bridge-dummy/src/main/kotlin/com/catenax/bpdm/bridge/dummy/service/PoolUpdateService.kt
```diff
@@ -53,16 +54,30 @@ class PoolUpdateService(
...
+                updateRequests.add(
+                    LegalEntityPartnerUpdateRequest(
+                        legalEntity = gateToPoolLegalEntity(it.legalEntity),
+                        legalAddress = gateToPoolLogisticAddress(it.legalAddress.address),
+                        bpnl = it.bpn!!
+                    )
+                )
+            } catch (e: BpdmNullMappingException) {
+                logger.info { "Error processing ${it.externalId}" }
+                val errorMapped = ErrorInfo(LegalEntityUpdateError.LegalEntityDuplicateIdentifier, "Error on mapping '${it.legalEntity}' message ${e.message}", it.bpn)
```
nicoprow, line 71: When I test locally I run into another error here because the error message is too long for the sharing state as this is a 255 VARCHAR.

nicoprow (approved):

Issue #955, opened 2024-05-29
Tax Jurisdiction Code Attribute for Physical Postal Address of Business Partner Data Management Gate and Pool APIs


Sebastian-Wurm:
### Description

As a data steward and indirect consumer of the BPDM Gate and Pool APIs, I want to include and manage the tax jurisdiction code for the physical postal address of each business partner record, so that I can ensure precise tax calculation, compliance with tax regulations, and accurate reporting based on the tax jurisdiction of each business partner's location in my business processes.

### Acceptance Criteria

- the Gate API (and its documentation) must be updated  to include the "taxJurisdictionCode" attribute for the physical postal address of a business partner in the request and response payloads for create, update, and retrieve operations.
- the Pool API (and its documentation) must be updated to include the "taxJurisdictionCode" attribute for the physical postal address of a legal entity, site and address in the request and response payloads for retrieve operations.
- the "taxJurisdictionCode" must be an alphanumeric string with no special length restriction (256 chars are OK) to accommodate various jurisdiction coding schemes.
- the "taxJurisdictionCode" is non-mandatory in both APIs (later in an operating environment it is ensured that the attribute will always be returned in the Gate output and Pool APIs for countries, where the tax jurisdiction code is legally mandatory)

### Additional Information

see also https://github.com/eclipse-tractusx/sig-release/issues/606

PR #972, opened 2024-06-20
feat: add taxJurisdictionCode field to physical address

nicoprow:
## Description

This pull request adds the new field 'taxJurisdictionCode' to the business partner's physical address

- changed API models
- added database column
- adapted the mappings
- enhanced the test data
- updated the api documentation

implements to #955

## Pre-review checks

### DIFF
--- a/docs/api/gate.json
+++ b/docs/api/gate.json
@@ -3,12 +3,20 @@
   "info": {
     "title": "Business Partner Data Management Gate",
     "description": "A gate for a member to share business partner data with CatenaX",
-    "version": "6.0.0"
+    "version": "6.1.0-SNAPSHOT"
   },
   "servers": [
     {
       "url": "http://localhost:8081",
       "description": "Generated server url"
+    }
+  ],
+  "security": [
+    {
+      "open_id_scheme": []
+    },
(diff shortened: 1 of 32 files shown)


docs/api/gate.json
```diff
@@ -3,14 +3,22 @@
   "info": {
     "title": "Business Partner Data Management Gate",
     "description": "A gate for a member to share business partner data with CatenaX",
-    "version": "6.0.0"
+    "version": "6.0.2-SNAPSHOT"
```
SujitMBRDI, line 6: Version is updated now, need to update docs of api for  each service or we shall prefer to adapt the same during feature freeze?

SujitMBRDI (commented):
All changes look good to me. Just need to make decision related to api docs version.

docs/api/gate.json
nicoprow: I will update the API docs with the newest SNAPSHOT version and then we can have a final look again

SujitMBRDI: Yeah, we can do that.

SujitMBRDI (approved):

PR #980, opened 2024-06-27
feat(bpdm-gate): updated upload partner process to only process site and address partners

SujitMBRDI:
## Description

In this pull request, we have modified the partner upload process in the bpdm-gate service to only create or update site and address business partner. 
Additionally, the legal entity BPN value will now be derived from the tenant's BPNL value.

## why
[700](https://github.com/eclipse-tractusx/sig-release/issues/700)

## Pre-review checks


nicoprow (changes requested):
Looks good except one doubt:

I think this algorithm needs to make sure that there definitely exists a BPNL to which the user can add records. This means to me that if no tenantBpnl exists it should return an error instead of trying to create records of an unknown legal entity.

What do you think?

@SujitMBRDI

SujitMBRDI:
> Looks good except one doubt:
> 
> I think this algorithm needs to make sure that there definitely exists a BPNL to which the user can add records. This means to me that if no tenantBpnl exists it should return an error instead of trying to create records of an unknown legal entity.
> 
> What do you think?
> 
> @SujitMBRDI

@nicoprow Yes, I agree to your suggestion here.
Even i think the existence of tenantBpnl must be first validation check for this algorithm before go into other field mapping process. If tenantBpnl is missing then ideally we should not perform upload partner upload process.

nicoprow (approved):

PR #981, opened 2024-07-02
feat(LegalSiteAPI): New endpoint to create legal site reference

dilipdhankecha2530:
## Description
What does this PR introduce? 
- Create an endpoint to create site entity
- Create SiteCreateRequestWithLegalAddressAsMain in bpdm-pool-api module
- Add test cases which cover the new code
- This PR solve the https://github.com/eclipse-tractusx/sig-release/issues/739.

Does it fix a bug? 
- Nope

Does it add a new feature?
- Yes

## Pre-review checks

### DIFF
--- a/bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/PoolSiteApi.kt
+++ b/bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/PoolSiteApi.kt
@@ -29,6 +29,7 @@ import io.swagger.v3.oas.annotations.tags.Tag
 import org.eclipse.tractusx.bpdm.common.dto.PageDto
 import org.eclipse.tractusx.bpdm.common.dto.PaginationRequest
 import org.eclipse.tractusx.bpdm.pool.api.PoolSiteApi.Companion.SITE_PATH
+import org.eclipse.tractusx.bpdm.pool.api.model.request.SiteCreateRequestWithLegalAddressAsMain
 import org.eclipse.tractusx.bpdm.pool.api.model.request.SitePartnerCreateRequest
 import org.eclipse.tractusx.bpdm.pool.api.model.request.SitePartnerUpdateRequest
 import org.eclipse.tractusx.bpdm.pool.api.model.request.SiteSearchRequest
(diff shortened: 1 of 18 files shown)


bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/PoolSiteApi.kt
```diff
@@ -133,4 +134,20 @@ interface PoolSiteApi {
         @ParameterObject searchRequest: SiteSearchRequest,
         @ParameterObject paginationRequest: PaginationRequest
     ): PageDto<SiteWithMainAddressVerboseDto>
+
+    @Operation(
+        summary = "Get page of sites matching the pagination search criteria",
+        description = "This endpoint retrieves all existing business partners of type sites."
```
nicoprow, line 140: The description does not match the endpoint behaviour

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/PoolSiteApi.kt
```diff
@@ -133,4 +134,20 @@ interface PoolSiteApi {
...
+    )
+    @ApiResponses(
+        value = [
+            ApiResponse(responseCode = "200", description = "Page of business partners matching the search criteria, may be empty"),
```
nicoprow, line 144: Also here

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/PoolSiteApi.kt
```diff
@@ -133,4 +134,20 @@ interface PoolSiteApi {
...
+            ApiResponse(responseCode = "400", description = "On malformed pagination request", content = [Content()])
+        ]
+    )
+    @Tag(name = ApiCommons.SITE_NAME, description = ApiCommons.SITE_DESCRIPTION)
+    @PostMapping("/legal/reference")
```
nicoprow, line 149: I'm not very happy with the resource name. Every site has a reference to a legal entity, which this name implies.

I propose /v6/legal-address-sites
or /v6/legal-main-sites

I'm open for suggestions but for me the name should be a resource name describing clearly a site that has a legal address as its main address, since this is the differentiating factor

nicoprow (changes requested):
This is a really good first pull request. The actual logic looks good to me and also tests are there.

I'm not happy with the naming of the REST resource though and this pull request would need to be rebased on the current main (during my review this could be done without conflicts).

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/PoolSiteApi.kt
dilipdhankecha2530: @nicoprow 

I added this logic to the PoolSiteApi, so /sites will follow v6. The endpoint will look like **/v6/sites/legal-address**.
Does it works?

nicoprow: To be honest I'm not really happy with that path because it implies that you want to create a legal address that belongs to a site. But in fact, you want to create a site with a special property - that is - it has the legal address as its main address.

I'm open to suggestions but from my perspective the very last name in the path should describe the name of the resource you want to create which will be a site with the special property of having the legal entity's legal address as its site main address.

dilipdhankecha2530: As per our discussion, i was update the endpoint like **/v6/sites/legal-main-sites**.

nicoprow (changes requested):
Content works and looks great! 

Last bit is about conformity:

- Please rebase your branch on the newest version of main
- Squash your commits to one so that only one commit is on top of the main branch
- This is not a patch version upgrade but since it brings anew feature to the API the maven version should be updated to 6.2.0-SNAPSHOT
- Please add an entry into the CHANGELOG.md file  under

```
## [6.2.0] - tbd
```

nicoprow (approved):

PR #984, opened 2024-07-03
build(charts): increase BPDM charts app-version to 6.0.2

nicoprow:
## Description

This pull request increases the BPDM charts app-version to 6.0.2

## Pre-review checks

### DIFF
--- a/charts/bpdm/CHANGELOG.md
+++ b/charts/bpdm/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 The format is based on Keep a Changelog (https://keepachangelog.com/en/1.0.0/),
 
+## [5.0.2] -  2024-07-03
+
+### Changed
+
+- Increase appversion to 6.0.2
+- update BPDM Pool Chart to version 7.0.2
+- update BPDM Gate Chart to version 6.0.2
+- update BPDM Orchestrator Chart to version 3.0.2
+- update BPDM Cleaning Service Dummy Chart to version 3.0.2
+
 ## [5.0.1] -  2024-05-27
 
 ### Changed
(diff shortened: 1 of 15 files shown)


nicoprow:
Linting workflow does not work correctly here as it uses the last tag in the repo instead of using the last tag on the branch to compare old and new versions.

SujitMBRDI:
> Linting workflow does not work correctly here as it uses the last tag in the repo instead of using the last tag on the branch to compare old and new versions.

Yeah, last and latest tag in our repo currently 5.1.0-SNAPSHOT but which is in main, we cannot bump it up from there.
can we go for 5.1.0 then? or do you see other option.

nicoprow:
> > Linting workflow does not work correctly here as it uses the last tag in the repo instead of using the last tag on the branch to compare old and new versions.
> 
> Yeah, last and latest tag in our repo currently 5.1.0-SNAPSHOT but which is in main, we cannot bump it up from there. can we go for 5.1.0 then? or do you see other option.

The workflow checks are just there to alert us of some problems that might be there in the pull request. If you can see as well that the version bump is correct and everything else about the charts is correct as well, you can approve anyway and we merge it

SujitMBRDI:
> > > Linting workflow does not work correctly here as it uses the last tag in the repo instead of using the last tag on the branch to compare old and new versions.
> > 
> > 
> > Yeah, last and latest tag in our repo currently 5.1.0-SNAPSHOT but which is in main, we cannot bump it up from there. can we go for 5.1.0 then? or do you see other option.
> 
> The workflow checks are just there to alert us of some problems that might be there in the pull request. If you can see as well that the version bump is correct and everything else about the charts is correct as well, you can approve anyway and we merge it

Got confused as pull request is in draft state, thought merging this changes will not possible unless we fix linting checks.
For patch version, i can see version bump for app and helm chart are correct.

SujitMBRDI (approved):

PR #986, opened 2024-07-03
feat(bpdm-gate): Adding a externalSequenceTimestamp attribute to the business partner input request

kunyao-cofinity-x:
…quest

## Description

The BPDM gate requires functionality to handle data updates from the customer side based on timestamps. This feature ensures that the data is only updated if the new data from the customer has a more recent timestamp than the existing data.

https://github.com/eclipse-tractusx/sig-release/issues/726

- Customer sends data to the gate system with a timestamp attribute (currentness).
- The gate system receives the data and invokes the GR process.
- The GR process compares the received timestamp with the stored timestamp.
- If the received timestamp is newer, the data is updated, and a success message is logged and sent to the customer.
- If the received timestamp is not newer, the update is rejected, and an informative message is logged and sent to the customer.

## Pre-review checks

Please ensure to do as many of the following checks as possible, before asking for a committer review:

- [x] Self-testing on local environment
- [x] Test cases creation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ For changes to the BPDM Helm charts please consult the [changelog](charts/bpdm/C
 - BPDM System Test: End-to-end test CI/CD workflow setup for golden record process. ([#1155](https://github.com/eclipse-tractusx/bpdm/issues/1155))
 - Apps : Enhanced dependency readiness checks with a scheduler to verify connections to required services every 30 seconds and during startup for Pool, Cleaning Service Dummy, and Gate service. ([#1161](https://github.com/eclipse-tractusx/bpdm/issues/1161))
 - BPDM Gate: Add consistency check with the golden record Pool making sure that referenced BPNs are still existing ([#1130](https://github.com/eclipse-tractusx/bpdm/issues/1130))
+- BPDM Gate: Add externalTimestamp to prevent old request overwrite newer request 
 
 ### Changed
 
@@ -114,15 +115,15 @@ For changes to the BPDM Helm charts please consult the [changelog](charts/bpdm/C
 
 ### Added
 
-- BPDM Gate: Configuration to prevent the uploaded business partner input data to immediately enter the golden record process. 
-In this configuration the business partner data needs to be sent to the golden record process manually over the new state/ready API endpoint.
-Default configuration remains automatically sharing.
+- BPDM Gate: Configuration to prevent the uploaded business partner input data to immediately enter the golden record process.
+  In this configuration the business partner data needs to be sent to the golden record process manually over the new state/ready API endpoint.
+  Default configuration remains automatically sharing.
 - BPDM Gate: Limited multi-tenancy support. Business partners are now separated by owner-BPNL.
-The owner is determined from the 'bpn' claim in the token.
-This means users of a Gate can only see and edit their own business partner data.
+  The owner is determined from the 'bpn' claim in the token.
+  This means users of a Gate can only see and edit their own business partner data.
 - BPDM Pool: New API endpoints to query business partner data which belongs to Catena-X members only
 - APIs: Added a major version number to all API endpoint paths indicating the current version of the BPDM APIs.
-In the future we will use version numbers in the URL to differentiate between all currently supported major versions of the API
+  In the future we will use version numbers in the URL to differentiate between all currently supported major versions of the API
 - BPDM Gate Client: Now supports the stats endpoints of the BPDM API
 
 ### Changed
@@ -134,15 +135,15 @@ In the future we will use version numbers in the URL to differentiate between al
 - BPDM Gate: Fix not correctly updating business partner output data from golden record updates in the Pool.
 - JAVA version to 21
 - BPDM API Permissions: Overhaul of the permissions needed to access the BPDM API endpoints.
-Permissions are now more fine-granular and differentiate more clearly between read or write.
-For more details consult the Arc42, API documentation and properties files of the respective applications.
+  Permissions are now more fine-granular and differentiate more clearly between read or write.
+  For more details consult the Arc42, API documentation and properties files of the respective applications.
 - BPDM App Configuration: Now all applications are secured (authenticated and authorized) by default.
-You can still deactivate security in the BPDM apps for testing or development purposes though.
+  You can still deactivate security in the BPDM apps for testing or development purposes though.
 - BPDM Pool: Fix Pool trying to update golden records which the golden record process indicated to have no changes
 - BPDM Orchestrator: The business partner data for golden record process tasks has been completely overhauled.
-Now business partner data is clearly divided into `uncategorized`, `legal entity`, `site` and `additonal address` data.
-This model is less verbose and contains less duplicate data.
-Additionally, both Pool and Gate can write and read from it making it unnecessary for a cleaning service to provide the data in two different models.
+  Now business partner data is clearly divided into `uncategorized`, `legal entity`, `site` and `additonal address` data.
+  This model is less verbose and contains less duplicate data.
+  Additionally, both Pool and Gate can write and read from it making it unnecessary for a cleaning service to provide the data in two different models.
 
 
 
@@ -190,7 +191,7 @@ Additionally, both Pool and Gate can write and read from it making it unnecessar
 - BPDM Gate: New business partner type 'GENERIC' for changelog
 - BPDM Gate: New business partner type 'GENERIC' for sharing state
 - Workflows: Trivy now targets the latest alpha Docker image instead of the latest release version
-- Apps: Increase projectreactor.netty version to fix Trivy vulnerability 
+- Apps: Increase projectreactor.netty version to fix Trivy vulnerability
 
 
 ## [4.0.1] - 2023-08-28
@@ -228,8 +229,8 @@ Please create a back-up of your business partner data before updating.
 - BPDM: Umbrella Chart with BPDM Bridge Dummy.
 
 ### Fixed
-  - BPDM: Deprecated endpoints for retrieving business partners in legacy format.
-  - Endpoint for retrieving changelog entries has now improved filtering (breaking API change)
+- BPDM: Deprecated endpoints for retrieving business partners in legacy format.
+- Endpoint for retrieving changelog entries has now improved filtering (breaking API change)
 
 ## [3.2.2] - 2023-05-12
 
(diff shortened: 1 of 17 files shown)


bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/generic/BusinessPartnerDb.kt
```diff
@@ -104,6 +105,9 @@ class BusinessPartnerDb(
     @JoinColumn(name = "address_confidence_id", unique = true)
     var addressConfidence: ConfidenceCriteriaDb?,
 
+    @Column(name = "currentness")
+    var currentness: Timestamp? = null,
```
SujitMBRDI, line 109: Suggestion from my side, since **java.time.Instant** is a more modern and efficient API for dealing with timestamps, it is advisable to switch to Instant. 
This also avoids potential issues with different time zone handling in java.sql.Timestamp.

Also, here [DbTimestamp](https://github.com/eclipse-tractusx/bpdm/blob/bedcf0703b9fa256de97d6bef2b3a0d88fce0570/bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/entity/DbTimestamp.kt#L30) wrapper for truncation can be created as we already did for orchestrator entities. This will ensure that timestamps are always truncated to microseconds, which can be useful if you want to avoid precision issues between application and the database.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -124,12 +124,21 @@ class BusinessPartnerService(
         val partnerToUpsert = existingPartner ?: BusinessPartnerDb.createEmpty(upsertData.sharingState, upsertData.stage)
 
         val hasChanges = compareUtil.hasChanges(upsertData, partnerToUpsert)
+        val shouldUpdate = if(upsertData.currentness == null) {
```
SujitMBRDI, line 127: Refactor the upsertFromEntity function: Improve readability and efficiency by avoiding redundant null checks and using more idiomatic Kotlin. We can have something like below,
```
val hasChanges = compareUtil.hasChanges(upsertData, partnerToUpsert)
val shouldUpdate = when {
       upsertData.currentness == null -> true
       existingPartner?.currentness == null -> true
       else -> upsertData.currentness.isAfter(existingPartner.currentness)
 }

 if (hasChanges && shouldUpdate) {
       changelogRepository.save(ChangelogEntryDb(sharingState.externalId, sharingState.tenantBpnl, changeType, stage))
       copyUtil.copyValues(upsertData, partnerToUpsert)
       businessPartnerRepository.save(partnerToUpsert)
}

```

SujitMBRDI (changes requested):
In general, Instant type can be used instead of Timestamp for better performance and modern API advantages, we can add DbTimestamp wrapper class to ensure truncation to microseconds and Also, refactor mapping and service function for improved readability and efficiency to ensure proper exception handling and null safety throughout the code

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/generic/BusinessPartnerDb.kt
SujitMBRDI: My recommendation is switch to `Instant` as is the best choice due to its modern design, better time zone handling, and higher precision. Handling null and string values can be easily managed with appropriate parsing and conversion logic.

Sebastian-Wurm (changes requested):
Please see my comment regarding currentness being a business attribute which should not be abused to solve technical problems.

nicoprow (changes requested):
Currently, the project does not compile on the test stage as there are some errors with the test objects.

As a first step I would recommend to clean up your commit history and reach a state in whcih you have one new commit that is on top of the current main branch. Then you can debug.

At the moment your commit history shows merges and and it references an outdated commit that is behind the current main.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/entity/generic/BusinessPartnerDb.kt
```diff
@@ -104,6 +105,9 @@ class BusinessPartnerDb(
     @JoinColumn(name = "address_confidence_id", unique = true)
     var addressConfidence: ConfidenceCriteriaDb?,
 
+    @Column(name = "currentness")
+    var currentness: Instant? = null,
```
SujitMBRDI, line 109: @kunyao-cofinity-x, need to cross check on requirement as do we really need timestamp on output stage? Also even if we add then can you please streamline attribute name as `externalSequenceTimestamp` only as it will create confuse in future while reading DB table content.

kunyao-cofinity-x: Hello @SujitMBRDI, thanks for looking into it. This attribute will be only applied to the input stage. I have pushed the modification of the column name.

SujitMBRDI:
Hello @kunyao-cofinity-x,
Cc: @nicoprow 

Code contribution looks fine but kindly perform below formalities to make your contribution available to main.
- [ ] rebase to the newest state of main
- [ ] update openAPI documents in the docs/api folder
- [ ] Add a CHANGELOG entry with a reference to the issue/PR
- [ ] Squash commits to have only 1 commit

kunyao-cofinity-x:
> Hello @kunyao-cofinity-x, Cc: @nicoprow
> 
> Code contribution looks fine but kindly perform below formalities to make your contribution available to main.
> 
> * [ ]  rebase to the newest state of main
> * [ ]  update openAPI documents in the docs/api folder
> * [ ]  Add a CHANGELOG entry with a reference to the issue/PR
> * [ ]  Squash commits to have only 1 commit

Hi @SujitMBRDI, the following points are done
- [x] rebase to the newest state of main
- [x] update openAPI documents in the docs/api folder
- [x] Add a CHANGELOG entry with a reference to the issue/PR
- [x] Squash commits to have only 1 commit
cc : @nicoprow

SujitMBRDI (approved):
LGTM

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
```diff
@@ -128,12 +128,17 @@ class BusinessPartnerService(
...
+        if (hasChanges && shouldUpdate) {
+                changelogRepository.save(ChangelogEntryDb(sharingState.externalId, sharingState.tenantBpnl, changeType, stage))
 
-            copyUtil.copyValues(upsertData, partnerToUpsert)
-            businessPartnerRepository.save(partnerToUpsert)
+                copyUtil.copyValues(upsertData, partnerToUpsert)
+                businessPartnerRepository.save(partnerToUpsert)
         }
 
         return UpsertResult(hasChanges, changeType, partnerToUpsert)
```
nicoprow, line 144: The upsert result should give the information whether the record did indeed change. Before there was only `hasChanges` as a criterion but now we have also `shouldUpdate`. In the UpsertResult we should therefore pass the result of both values.

nicoprow (changes requested):
Just found a problem upon close inspection with the UpsertResult. See below for more details. 

The current behaviour prevents the record from being updated but still sets the sharing state to `Initial` even though the business partner data has not been updated

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerService.kt
kunyao-cofinity-x: 1. extend a new attribute to UpsertResult
2. return UpsertResult(hasChanges, **shouldUpdate**, changeType, partnerToUpsert)
3. .takeIf { it.hadChanges || **it.shouldUpdate** || sharingState.sharingStateType == SharingStateType.Error }

nicoprow (approved):

nicoprow (changes requested):
I found a case when the logic does not work as expected:

1. Create and share abusiness partner with an external timestamp A
2. Watch it go to Pending State
3. Update the business partner with an earlier timestamp A-1 
4. The record data is not updated as expected but
5. Look into the sharing state and see that it got into ready again even though the data has not been updated

This is because the check for the upserted data is wrong. I recommend to delete the two flags in upsertedData class and replace it with something like "hasBeenUpserted" or "dataHasChanged" to indicate that the data has been successfully changed. The check is the result of the two conditions hasChanges and shouldUpdate. Querying that flag then when determing whether to update the sharing state should lead to the right behaviour

nicoprow (approved):
I tested different scenarios and everything seems to work fine:

1. Create business partner without external sequence timestamp -> Creation successful
2. Update business partner without extenral sequence timestamp -> Update successful
3. Update with no changes to business partner without external sequence timestamp -> No Update
4. Create business partner with external sequence timestamp and share -> Creation successful
5. Update business partner with newer sequence timestamp -> Update successful
6. Update business partner with older sequence timestamp -> No update
7. Update business partner with newer sequence timestamp but no changed data -> No update
8. Update business partner in error state with older sequence timestamp but no changed data -> No Update
9. Update business partner in error state with newer sequence timestamp but no changed data -> Update successful (start sharing again)

CHANGELOG.md
```diff
@@ -112,15 +112,15 @@ For changes to the BPDM Helm charts please consult the [changelog](charts/bpdm/C
```
SujitMBRDI, line 112: I couldn't see the changelog entry added for this feature.
As app version gonna be is 6.3.0 Kindly add changelog entry for the same.

SujitMBRDI (changes requested):
The code changes looks good to me!
I have performed below tests -
1. Created business partner with and without externalSequenceTimestamp attribute
2. Tried updating business partner record with and without externalSequenceTimestamp
3. Also performed check while updating record with older timestamp version of externalSequenceTimestamp 

All scenario worked as expected. 🚀 Only change expected in CHANGELOG.md file.
Also please rebase branch to main and have squashed commits to the pull request then it will be good for merge.

CHANGELOG.md
kunyao-cofinity-x: Hello @SujitMBRDI, I have added this implementation into the Add section in the changelog

SujitMBRDI (approved):

PR #988, opened 2024-07-05
fix(gate): Starting Golden Record Process Without Changes

nicoprow:
## Description

When upserting business partner input to the BPDM Gate that matches the existing business partner input, the Gate would still send the unchanged input data unnecessarily to the Golden Record Process. 

This pull request fixes this behaviour. Now the Gate only sends data to the Golden Record Process if it has been changed.

## Pre-review checks


SujitMBRDI (approved):
looks good to me

PR #1097, opened 2024-10-29
fix(bpdm-gate): output stage to only provide states based on business partner type

SujitMBRDI:
## Description

In this pull request, we have fixed logic for gate output stage to retrieve states based on business partner type.

## Pre-review checks

### DIFF
--- a/bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerMappings.kt
+++ b/bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerMappings.kt
@@ -65,7 +65,15 @@ class BusinessPartnerMappings {
                     } }
                     .map(::toIdentifierDto)
             }?: emptyList(),
-            states = entity.states.map(::toStateDto),
+            states = entity.postalAddress.addressType?.let { addressType ->
+                entity.states
+                    .filter { it.businessPartnerTyp == when (addressType) {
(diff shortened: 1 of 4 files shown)


bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerMappings.kt
```diff
@@ -65,7 +65,7 @@ class BusinessPartnerMappings {
                     } }
                     .map(::toIdentifierDto)
             }?: emptyList(),
-            states = entity.states.map(::toStateDto),
+            states = toStateDtos(entity.states, BusinessPartnerType.GENERIC),
```
nicoprow, line 68: I'm just looking at the identifiers behaviour. Shouldn't the states not better mirror the behaviour of the identifiers and take the states based on the address type?

nicoprow (commented):
Got some doubts we should clarify

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/BusinessPartnerMappings.kt
SujitMBRDI: In my opinion, yes we can do that and it will be right approach as well.

SujitMBRDI:
> Got some doubts we should clarify

I have adapted changes as we discussed.

nicoprow (approved):
Looks good to me

Issue #1082, opened 2024-10-17
QG checks [BPDM] (Release 24.12)


SujitMBRDI:
## QG checks

Please open and fill in this issue in your product repository to document the compliance with our Tractus-X Release Guideline (TRGs)

**Show compliance with TRGs** by referencing to a tagged link in the respective repository where possible, example: TRG 1.01 (see [github.com/eclipse-tractusx/example-repo/tree/1.0.0/README.md](https://github.com/eclipse-tractusx/example-repo/tree/1.0.0/README.md))

### Close this issue once the compliance with the TRGs has been documented

Committer(s): @nicoprow, @SujitMBRDI 
Helm Chart Version: 5.2.0
App Version: 6.2.0

Release Management Reference Issue: 

### Check of Tractus-X Release Guidelines

- Currently implemented automatic checks can be found under your product on our [Release Guidelines Checks Board](https://eclipse-tractusx.github.io/sig-release/)
- This QG Check is depending on the mandatory information from our current [Release Guidelines](https://eclipse-tractusx.github.io/docs/release)

#### TRG 1 Documentation

- [x] [TRG 1.01](https://eclipse-tractusx.github.io/docs/release/trg-1/trg-1-1) appropriate `README.md`
- [x] [TRG 1.02](https://eclipse-tractusx.github.io/docs/release/trg-1/trg-1-2) appropriate install instructions either `INSTALL.md` or in `README.md`
- [x] [TRG 1.03](https://eclipse-tractusx.github.io/docs/release/trg-1/trg-1-3) appropriate `CHANGELOG.md`
- [x] [TRG 1.04](https://eclipse-tractusx.github.io/docs/release/trg-1/trg-1-4) editable static files
- [x] [TRG 1.05](https://eclipse-tractusx.github.io/docs/release/trg-1/trg-1-05) architecture docs
- [x] [TRG 1.06](https://eclipse-tractusx.github.io/docs/release/trg-1/trg-1-06) administrator guide
- [x] [TRG 1.07](https://eclipse-tractusx.github.io/docs/release/trg-1/trg-1-07) user manual
- [x] [TRG 1.08](https://eclipse-tractusx.github.io/docs/release/trg-1/trg-1-08) open api docs

#### TRG 2 Git

- [x] [TRG 2.01](https://eclipse-tractusx.github.io/docs/release/trg-2/trg-2-1) default branch is named `main`
[...]

PR #1132, opened 2024-11-26
docs(charts): update helm chart README for app 6.2.0

SujitMBRDI:
## Description

In this pull request, we have updated README file for umbrella and for all sub charts along with changelog for added changes for version 5.2.0.

contributes to https://github.com/eclipse-tractusx/bpdm/issues/1082

## Pre-review checks


nicoprow (changes requested):
Nice to see that our checks are useful and catch some minor errors ^^
If we change the helm chart we need to either upgrade it to the release version for put it back to snapshot. Otherwise it will be released again as a release candidate...

SujitMBRDI:
> Nice to see that our checks are useful and catch some minor errors ^^ If we change the helm chart we need to either upgrade it to the release version for put it back to snapshot. Otherwise it will be released again as a release candidate...

Yes 😀 our checks are useful.
I later realise the same as we should not update docs may be for release candidate for now. While drafting final release we can update it.

nicoprow (approved):
LGTM

PR #1156, opened 2024-12-10
feat: handle priority for golden records from the orchestrator

dilipdhankecha2530:
## Description

This PR implements priority handling (High, Low) based on thresholds within the orchestrator. It also adds new endpoints for registering gate components with priorities. The orchestrator will generate the `OriginId` automatically and that will be used by the gate component.

Issue: https://github.com/eclipse-tractusx/sig-release/issues/856

## Pre-review checks


dilipdhankecha2530:
@nicoprow @SujitMBRDI Raise the PR just for the feedback purpose. I am still working on it.

nicoprow (commented):
I think the solution with an OriginID is problematic here.

1. Can we not replace the origin-ID with the the tenant BPNL. Each Gate knows already its tenant and even can be multi-tenant in the case of the Portal. This way we do not need to introduce another identifier
2. Now, each task is assigned with the origin-ID in the Orchestrator. This makes it exoplicit which tasks belong to the same company and also makes it easy to identify which company it is (for example based on the own company data). The Tasks should be assigned with priority and not with the origin-ID directly

nicoprow:
I removed this pull request from the 6.3.0 milestone

Issue #1155, opened 2024-12-10
Create End-to-End test workflow


SujitMBRDI:
### Description
BPDM project now contains `bpdm-system-tester` module which have capability to perform E2E test for golden record processes.
This issue adds feature to automate execution for bpdm-system-tester application using workflows.

### Acceptance Criteria

- [x] Create docker image for bpdm-system-tester
- [x] Add chart test to perform E2E testing

### Additional Information

Once we add workflow to the project then E2E test will be performed for every new app version.

SujitMBRDI:
Closed https://github.com/eclipse-tractusx/bpdm/issues/1157 as completed.

PR #1157, opened 2024-12-10
feat(bpdm-system-test): create end to end test CI/CD workflow for golden record process

SujitMBRDI:
## Description
This pull request adds CI/CD workflow for golden record process. 

-  Dockerization of bpdm-system-tester module
-  Chart test enabled 
-  Made Postgres init script executable by changing mode from 100644 → 100755

Contributes to #1155 

## Pre-review checks


nicoprow (commented):
@SujitMBRDI 
In order for the Helm test to work I recommend using the Central-IDP 4.0.1 as the realm seeding happens much faster there. It should be finished before the other components have been implemented. With that change it could also work without out the BPDM dependency check.

SujitMBRDI:
> @SujitMBRDI In order for the Helm test to work I recommend using the Central-IDP 4.0.1 as the realm seeding happens much faster there. It should be finished before the other components have been implemented. With that change it could also work without out the BPDM dependency check.

Thanks for the help @nicoprow.
I have updated delay timeout for steps here as 4 minutes.

nicoprow (approved):
LGTM

Issue #1027, opened 2024-08-09
feat(bpdm-gate-pool): MVP hierarchy data model adjustment


HeyHardy:
@nicoprow EDIT:

### Use case Description

As a Company Admin,
I want to manage hierarchies for Business Partner Number Legal Entities (BPNLs) in the Catena-X Portal,
so that I can effectively organize and control the relationships between different BPNLs by using the existing UI for own company data in the portal.

### Acceptance Criteria

All acceptance criteria refer to the BPDM Gate component. Other BPDM components are not affected as the business partner relationship will not be shared for now.

1. **Data Modeling and Integration:**
   - [x] Create a new business partner relationship resource representing the hierarchy (see chapter `Relationship Abstract Data Model`).
   - [x] Support relationship type 'IsManagedBy'.
   - [x] Ensure that each BPNL can support multiple hierarchy types.

2. **CRUD Operations:**
   - [x] **Create:** Ability to create a new **business partner relationship** with hierarchy type `IsManagedBy`
   - [x] **Read:** Display all hierarchy types, including `Managed in Portal` with their details.
   - [x] **Update:** Allow editing of existing hierarchies.

3. **Extended Functional Requirements**
   - [x] Adjust authorization logic
   - [x] Only able to define relationship for own Legal Entity (identified by owner BPNL) as source and other owned Legal Entity as target
   - [x] No duplicate relationships from same source to same type

### Relationship Abstract Data Model

```mermaid
---
title: Business Partner Relationship
---
classDiagram
    BusinessPartnerRelationship "*" --> "1" BusinessPartner: Source
    BusinessPartnerRelationship "*" --> "1" BusinessPartner: Target
    BusinessPartnerRelationship "*" --> "1" BusinessPartnerRelationshipType: Type

    class BusinessPartner["Business Partner"]{
    }
    class BusinessPartnerRelationship["Business Partner Relationship"]{
        String ExternalId
        Timestamp CreatedAt
        Timestamp UpdatedAt 
    }
    class BusinessPartnerRelationshipType["Business Partner Relationship Type"]{
        <<Enumeration>>
        IsManagedBy
    }
```

#### Permissions

1. read_input_relationship
2. write_input_relationship

#### BPDM Roles

<table>
  <tbody>
    <tr>
      <th>Admin</th>
      <th>Input Manager</th>
      <th>Input Consumer</th>
    </tr>
    <tr>
      <td>
          <li>read_input_relationship</li>
[...]

PR #1167, opened 2024-12-19
feat(Gate): add business partner relations

nicoprow:
## Description

This pull request adds  new endpoints to search, create, update and delete business partner relations.
It currently supports only the IsManagedBy relation type.

- add logic for CRUD and constraint checking
- add database table for relations
- add tests

Contributes to #1027

## Pre-review checks


SujitMBRDI (changes requested):
Great logic and code changes!!
Changes looks good to me, tests i have performed: 
1. From gate input, I have provided two legal entities 
2. Then created relations using tenant null
3. Then deleted relationship too.

As we discussed in review session, just we need to swap logic for `Source` and `Target` as `IsManagedBy` is passive term.

nicoprow:
I changed the constraints on the IsManagedBy relation as discussed.
Also, I removed the tenantBpnL as it caused confusion before. Maybe at some later point we can add it again as a hidden parameter that only appears on no-auth profiles. But that is quite extra work that I feel warrants its own contribution.

SujitMBRDI (approved):

PR #1173, opened 2025-01-09
fix(Pool): duplicate BPN request identifier mapping entries

### LINKED ISSUE #1159: Duplicate entries in bpn_requestIdentifier_mapping
### Current Behavior
- Duplicate mapping entries for each relation between requestIdentifier and BPN

### Expected Behavior
- Singular entries for each relation between requestIdentifier and BPN
- Duplicate entries are removed from the bpn_requestIdentifier_mapping table

### Steps To Reproduce
1. Process multiple different tasks with the same requestIdentifier

nicoprow:
## Description

This pull request fixes the Pool's behaviour of creating duplicate BPN request identifier mapping entries

- Remove duplicate BPN request identifier mappings form the database
- Prevent creation of new duplicates
- Fix the request identifier unique constraint so that an exception will be thrown when a duplicate entry tries to be created in the database (just for data integrity, should not happen in normal logic)

Fixes #1159

## How to test

In order to check the behaviour you can create a business partner with site information via the Gate, full golden record process. Then provide an update to that business partner without changing the names or BPNs (for example change a small field in the address). Look into the database, there should be no duplicate entries. In the previous version you should see a duplicate.

Please fill free to execute further test scenarios.

## Pre-review checks


SujitMBRDI (approved):
Looks good to me, 
The tests i have performed where i tried to perform round trip by creating Legal Entity with same identifier and got expected error result.
Also on Pool, tried creating legal entity with same identifier and worked as expected.

Issue #1130, opened 2024-11-26
Consistency check between Gate and Pool


StWeisshaar:
### Description
As user of BPDM,
I would like that the BPNs shown in the Gate Output Stage are consistent with the BPNs stored in the Pool,
so that the information stored in the Gate can be trusted.

### Acceptance Criteria

- [x] Consistency check between Gate Output Stage and Pool
- [x] Error sharing state type to inform users about inconsistencies
- [x] Configurable schedule for the consistency check

### Additional Information
There are currently no endpoints to remove issued BPNs from the Pool. In the case of duplicates we must delete them directly from the Pool DB. Making sure that these BPNs are not only removed from the Pool but also from the Gates is error prone and should be supported by a consistency check from the Gate. Especially for de-centralized Gates it makes sense to have a consistency check in place that can validate the stored BPNs. We must make sure that this feature does not open a backdoor to fetch information about all existing BPNs.

nicoprow:
@StWeisshaar I assign you this issue to drive it to a refined state. Let's talk about it in the next refinement meeting and see if we can already put it to the refined state. For now , I have no more questions towards that topic and I could see it being worked on.

@SujitMBRDI what is your opinion?

nicoprow:
Note: there is already a sharing state error called 'BpnNotInPool'. I would suggest to use that one.

SujitMBRDI:
> @StWeisshaar I assign you this issue to drive it to a refined state. Let's talk about it in the next refinement meeting and see if we can already put it to the refined state. For now , I have no more questions towards that topic and I could see it being worked on.
> 
> @SujitMBRDI what is your opinion?

Yes, I can agree for the two topics in acceptance criteria those are as below.
1. Consistency check between Gate Output Stage and Pool -> No doubts, development wise it is clear.
2. Configurable schedule for the consistency check -> No doubts.

But for the topic `Error sharing state type to inform users about inconsistencies`, we have already sharing states error which  are defining inconsistencies e.g. `BpnErrorNotFound` and `BpnErrorTooManyOptions`.  Also, we can configure `BpnNotInPool` to make it more consistent. 
or are we expecting separate sharing state api endpoint where user(sharing member) can feed in BPNs for which consistency should be checked?

nicoprow:
Test Case

For each business partner type:

- Legal Entity
- Site
- Address

Gate with owner BPNL {{OWNER_BPNL}}

Gate POST input/business-partners:

```json
[
  {
[...]

PR #1190, opened 2025-01-21
feat(Gate): Consistency Check with the Pool 

nicoprow:
## Description

This pull request introduces a consistency check to the Gate which checks whether referenced BPNs in the output stage are still existing in the Pool. If a discrepancy has been found it will be written as an error in the sharing state for the business partner record in question.

For testing I recommend the following:

1. setup the full golden record process
2. Share some business partners linking to different golden records
3. Set the concistency check to a fairly regular interval in the Gate's configureation
4. See in the logs whether the consistency checks works without issues
5. Go to the Pool database and rename some linked BPNs
6. After the next consistency check is done check the sharing states of affected business partner records
7. You should see the errors in the sharing states

I did not add any automatic integration as the setup of these tests via mocking is very time consuming. We should decide whether we want to invest the time into actually creating such tests or keep those as manual tests.

Implements #1130 
## Pre-review checks


SujitMBRDI (approved):
Changes LGTM 🚀
The tests I have performed -
1. Reduced scheduling time for consistency  to 10seconds
2. Created golden record through round trip
3. Changed BPN number from POOL database
4. And checked scheduling - which worked as expected
5. Also got sharing state error with error code - BpnNotInPool

Regarding automated test, yes I agree that this will be time consuming thing and we'll achieve sometime later.

PR #1192, opened 2025-01-23
fix(bpdm-gate): update outdated task result to latest stage

### LINKED ISSUE #1185: Gate Not Updating Outdated Task Result
### Current Behavior

When uploading two business partner records that are matched to the same golden record in quick succession to each other,
the first business partner output to be retrieved is already outdated.
In this case the business partner output may then stay outdated as the golden record changelog has been missed.

### Expected Behavior

Even when an outdated business partner output is written to the Gate the Gate should make sure to get the newest update for that record.

### Steps To Reproduce

See current behaviour: Upload two record in quick succession. However, due to scheduling timings between the changelog check and the Gate's task resolution this effect does not always happen.

SujitMBRDI:
## Description

This pull request fixes the Gate not updating business partner output data when it receives already outdated output data form the golden record process. Now, when the Gate receives the result of golden record tasks it performs a check against the Pool again and may update data if it is already outdated.

Fixes #1185 

## Pre-review checks


SujitMBRDI:
For testing I recommend the following way:

1. setup the full golden record process
2. Insert one business partner record and let it perform round trip.
3. Upload two or more business partner records with updated data that are matched to the same golden record in quick succession to each other.
4. After round trip, you should see each every record is updated to latest stage.

I did not had any integrated test for this scenario as the setup of these tests via mocking is very time consuming. 
But there is scope where we can have one E2E test that will cover automated test for this bug.

SujitMBRDI:
@dilipdhankecha2530, could you please review as Nico is on vacation and we are already aligned on this topic.

dilipdhankecha2530 (approved):
LGTM

PR #1193, opened 2025-01-24
Fix/outdated resolved tasks

### LINKED ISSUE #1185: Gate Not Updating Outdated Task Result

nicoprow:
## Description

This pull request fixes the Gate not updating business partner output data when it receives already outdated output data form the golden record process. Now, when the Gate receives the result of golden record tasks it performs a check against the Pool again and may update data if it is already outdated.

This pull request needs the fix for the Pool returning full task results based on the database state in order to work.

Fixes #1185 

## Pre-review checks


SujitMBRDI:
Closing this draft pull request, as adapted changes in another PR [1192](https://github.com/eclipse-tractusx/bpdm/pull/1192)

PR #1231, opened 2025-02-20
fix(Pool): not checking the LSA parent hierarchy

### LINKED ISSUE #1230: Wrong LSA Parents in Gate Output
### Current Behavior

As a sharing member when setting a new site `B` for an existing additional address of site `A`,  
but the golden record does not overtake that change for the additional address from site `A` to site `B`  
then the output incorrectly shows that the additional address now belongs to site `B` even though in the Pool the address actually still belongs to site `A`.

### Expected Behavior

I want to the output to actually be what is the correct state in the Pool, namely that the additional address still belongs to site `A`.

### Steps To Reproduce

1. In the Gate create an additional address of a site `A`
2. Wait for the golden record process to give you a BPNA
3. Update the additional address by BPNA reference and specify a new site  `B` that is not `A`
4. Wait for the golden record process to finish
5. The output for that updated record now shows its site `B`
6. Check the BPNs in the Pool and you notice that the actual site of the additional address is still `A` though

nicoprow:
## Description

This pull request introduces a validation check for incoming golden record tasks making sure that the Legal Entity, Site and Additional Address hierarchy is consistent in the contained business partner object. Since the Pool does not support changing the parent legal entity or site at the moment, it returns a validation error on encountering such golden record tasks. 

With this, it is up to the cleaning service to make sure that no such invalid business partner reaches the Pool. Otherwise the task will be returned to the sender with a sharing error.

Especially in regards to #1230 . This means, the Pool can also not return inconsistent data like that as it does not allow it to be processed.

Fixes #1230 

## Pre-review checks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ For changes to the BPDM Helm charts please consult the [changelog](charts/bpdm/C
 - BPDM Pool: Fix resolved tasks by providing business partner data based on the database state. Returned result is now consistent. ([#1158](https://github.com/eclipse-tractusx/bpdm/issues/1158))
 - BPDM Pool: Fix missing legal entity (BPNL) associated to address in response while performing entity search ([#1191](https://github.com/eclipse-tractusx/bpdm/issues/1191))
 - BPDM Pool: Fix not returning the most-up-to-date legal and site main address when it has been modified by an update to the site ([#1209](https://github.com/eclipse-tractusx/bpdm/issues/1209))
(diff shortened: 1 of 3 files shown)


SujitMBRDI (approved):
Changes looks good to me 🚀
Tests I performed - 
1. Created gate input request where provided business partner data like legal entity with legal address, site with site main address and additional address for both site and legal entity.
2. Wait for golden record process
3. Performed similar gate input request for new legal enitity and site but there provided old references(which i received from previous request) of additional address(with BPNs).
4. wait for golden record process
5. received sharing state error business partner where i have provided old references.

PR #1266, opened 2025-03-19
cicd: remove the license-check plugin from any maven phase

nicoprow:
## Description

Remove the license-check plugin from any maven phase.

This way the license-check is only executed when explicitly started by the goal and not during the maven lifecycle.

This solves the problem of the license-check interrupting the build process due to a too-many-requests error.

## Pre-review checks


SujitMBRDI (approved):
LGTM

PR #1296, opened 2025-03-31
cicd(Workflows): change trivy exit code

nicoprow:
## Description

The regular trivy workflow check on main currently breaks when finding a vulnerabilty. This indicates that something is wrong with the workflow or some critical part of the cicd is not working correctly. This however, not the case as trivy just reports some found vulnerabilties. 

I propose we change the exit code for this trivy check to 0 making sure that the github workflow does not break but still reports any found vulnerabilities in the security section of github. 

The behaviour of the trivy check during the pull request which breaks on finding a vulnerabilty remains unchanged.

## Pre-review checks


SujitMBRDI (approved):
LGTM

Issue #1294, opened 2025-03-31
Increase  API Version of BPDM of V7


nicoprow:
### Description

As a consumer of the BPDM APIs I want to slowly integrate breaking changes while still having a stable old version backwards compatible. The old and new API version should work interchangeably with one another for the duration that the old version is supported. The old version should not change in structure and behaviour for this duration.

### Acceptance Criteria

- Create a new version 7 of the API next to version 6
- Make sure version 7 and version 6 work interchangeable with one another
- Create new Version 7 assets

### Task Progress

- [x] Orchestrator API
- [x] Pool API
- [x] Gate API
- [x] Assets

nicoprow:
this issue has no sig-release parent. Such a parent should be created

PR #1308, opened 2025-04-08
feat(bpdm-gate): api version increased with simultaneous v6 support

SujitMBRDI:
## Description

This pull request adds new api version for BPDM Gate service.
The new api version is now v7 along with simultaneous support for previous version v6.
Currently, bpdm-gate service will support for both versioned apis but for round trip test we have enabled only v7.

Contributes to https://github.com/eclipse-tractusx/bpdm/issues/1294

## Pre-review checks


bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/RelationService.kt
```diff
@@ -103,6 +100,27 @@ class RelationService(
...
+            val existingRelationship = relationRepository.findByTenantBpnLAndExternalId(tenantBpnL.value, externalId)
+            if (existingRelationship != null) throw BpdmRelationAlreadyExistsException(externalId)
+        }
+        return toDto(createInputStage(tenantBpnL, externalId, relationType, sourceBusinessPartnerExternalId, targetBusinessPartnerExternalId))
+    }
+
+    @Transactional
+    fun deleteRelation(tenantBpnL: BpnLString, externalId: String) {
+        val relations = relationRepository.findByTenantBpnLAndExternalId(tenantBpnL.value, externalId) ?: throw BpdmMissingRelationException(externalId)
+        relationRepository.delete(relations)
```
nicoprow, line 121: The delete does not work as there is an input stage still attached to the relation.

First the input stage needs to be deleted and then afterwards the relation itself.

nicoprow (changes requested):
Apart from the delete functionality everyting else looks great.

bpdm-gate/src/main/kotlin/org/eclipse/tractusx/bpdm/gate/service/RelationService.kt
SujitMBRDI: Done.

nicoprow (approved):
LGTM

PR #1314, opened 2025-04-18
feat(Gate): add sharing state type and updatedAfter filter

nicoprow:
## Pre-review checks


SujitMBRDI (changes requested):
Code changes looks good to me.
Only thing missing here is to update api doc for bpdm-gate service.

nicoprow:
> Code changes looks good to me. Only thing missing here is to update api doc for bpdm-gate service.

good catch! I have updated the API docs as well now.

It seems like the Gate.yaml was in a different format before, so there is a lot of lines of code changes. But I took the generated yaml file from springdoc, so it should be in the correct format.

SujitMBRDI:
> > Code changes looks good to me. Only thing missing here is to update api doc for bpdm-gate service.
> 
> good catch! I have updated the API docs as well now.
> 
> It seems like the Gate.yaml was in a different format before, so there is a lot of lines of code changes. But I took the generated yaml file from springdoc, so it should be in the correct format.

Yeah, swagger editor now updated and started accepting json file formate by default as well. And there is no need to covert json to yaml. The reason why we are seeing many changes on gate.yaml  and correct format too.

SujitMBRDI (approved):

PR #1360, opened 2025-06-03
chore(app): added code coverage check using JaCoCo

SujitMBRDI:
## Description

This pull request adds proposal for having code coverage using JaCoCo during the build process for BPDM application.
Also, coverage reports and code quality metrics will be analyzed and visualized through SonarQube. Currently the BPDM  does not have an integration in this SonarQube instance which is need to fix first. 

Contributes to #1338

## Pre-review checks


nicoprow (approved):
LGTM

PR #1382, opened 2025-06-30
feat: Implement search logic for business partner extension

kunyao-cofinity-x:
## Description

- What does this PR introduce?

This feature aims to introduce enhanced search functionality for Business Partners in the CX network. It improves the current available options for identifying Business Partners in different scenarios like e.g., within the CX-onboarding process.
The benefit of the new functionality is that a member can search for a business partner based on several characteristics and receives the latest information out of the network.

- Does it add a new feature?

It is a new feature, but it is partially based on an existing data model.

## Pre-review checks

### DIFF
--- a/bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/request/LegalEntityPropertiesSearchRequest.kt
+++ b/bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/request/LegalEntityPropertiesSearchRequest.kt
@@ -26,9 +26,25 @@ import io.swagger.v3.oas.annotations.media.Schema
 @Schema(description = "Contains keywords used for searching in legal entity properties")
 data class LegalEntityPropertiesSearchRequest constructor(
     @field:Parameter(description = "Filter legal entities by name")
-    val legalName: String?
+    val legalName: String?,
+
+    @field:Parameter(description = "Filter business partners by CX-BPN.")
+    val id: String?,
+
+    @field:Parameter(description = "Filter business partners by street name.")
+    val street: String?,
+
+    @field:Parameter(description = "Filter business partners by zip code.")
+    val postcode: String?,
+
+    @field:Parameter(description = "Filter business partners by city.")
+    val city: String?,
+
+    @field:Parameter(description = "Filter business partners by country code ISO 3166-1.")
+    val country: String?,
+
 ) {
     companion object {
-        val EmptySearchRequest = LegalEntityPropertiesSearchRequest(null)
+        val EmptySearchRequest = LegalEntityPropertiesSearchRequest(null, null, null, null, null, null)
     }
 }
--- a/bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/response/BusinessPartnerSearchResultDto.kt
+++ b/bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/response/BusinessPartnerSearchResultDto.kt
@@ -1,0 +1,52 @@
+/*******************************************************************************
+ * Copyright (c) 2021,2024 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License, Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ******************************************************************************/
+
+package org.eclipse.tractusx.bpdm.pool.api.model.response
+
+import io.swagger.v3.oas.annotations.media.Schema
+import org.eclipse.tractusx.bpdm.pool.api.model.LegalEntityIdentifierDto
+import org.eclipse.tractusx.bpdm.pool.api.model.LegalFormDto
+import org.eclipse.tractusx.bpdm.pool.api.model.StreetDto
+
+@Schema(name = "BusinessPartnerSearchResultDto", description = "")
+data class BusinessPartnerSearchResultDto(
+    @Schema(description = "BPN L/S/A")
+    val id: String,
+
+    @Schema(description = "BPN L/S/A")
+    val name: String?,
+
+    @Schema(description = "Legal Form")
+    val legalForm: LegalFormDto?,
+
+    @Schema(description = "Street")
+    val street: StreetDto?,
+
+    @Schema(description = "City")
+    val city: String,
+
+    @Schema(description = "PostalCode")
+    val postalCode: String,
+
+    @Schema(description = "Country")
+    val country: String,
+
+    @Schema(description = "")
+    val identifiers: List<LegalEntityIdentifierDto>
+)
--- a/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/repository/LogisticAddressRepository.kt
+++ b/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/repository/LogisticAddressRepository.kt
@@ -19,9 +19,13 @@
 
 package org.eclipse.tractusx.bpdm.pool.repository
 
+import org.eclipse.tractusx.bpdm.pool.api.model.request.LegalEntityPropertiesSearchRequest
 import org.eclipse.tractusx.bpdm.pool.entity.LegalEntityDb
 import org.eclipse.tractusx.bpdm.pool.entity.LogisticAddressDb
+import org.eclipse.tractusx.bpdm.pool.entity.NameDb
+import org.eclipse.tractusx.bpdm.pool.entity.PhysicalPostalAddressDb
 import org.eclipse.tractusx.bpdm.pool.entity.SiteDb
+import org.eclipse.tractusx.bpdm.pool.entity.StreetDb
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.Pageable
 import org.springframework.data.jpa.domain.Specification
@@ -70,6 +74,73 @@ interface LogisticAddressRepository : JpaRepository<LogisticAddressDb, Long>, Jp
                     )
                 }
             }
+        /*
+        business-partner/search
+        */
+        fun buildAddressSpecification(request: LegalEntityPropertiesSearchRequest): Specification<LogisticAddressDb> {
+
+            return Specification.where(byBpnA(request.id))
+                .and(byStreet(request.street))
+                .and(byZipCode(request.postcode))
+                .and(byCity(request.city))
+                .and(byCountry(request.country))
+        }
+
+        private fun byBpnA(bpna: String?) =
+            Specification<LogisticAddressDb> { root, _, builder ->
+                bpna?.takeIf { it.isNotEmpty() }?.let {
+                    builder.equal(root.get<String>(LogisticAddressDb::bpn.name), bpna)
+                }
+            }
+
+        private fun byStreet(street: String?) =
+            Specification<LogisticAddressDb> { root, _, builder ->
+                street?.takeIf { it.isNotBlank() }?.let {
+                    builder.equal(root.get<LogisticAddressDb>("physicalPostalAddress")
+                        .get<StreetDb>("street")
+                        .get<String>("name"),
+                        "%${it.lowercase()}%")
+                }
+            }
+
+        private fun byZipCode(zipCode: String?) =
+            Specification<LogisticAddressDb> { root, _, builder ->
+                zipCode?.takeIf { it.isNotBlank() }?.let {
+                    builder.equal(root.get<PhysicalPostalAddressDb>("physicalPostalAddress").get<String>("postCode"), "%${it.lowercase()}%")
+                }
+            }
+
+        private fun byCity(city: String?) =
+            Specification<LogisticAddressDb> { root, _, builder ->
+                city?.takeIf { it.isNotBlank() }?.let {
+                    val normalizedVariants = normalizeGermanUmlauts(city)
+                    val joinAddress = root.join<LegalEntityDb, LogisticAddressDb>("addresses")
+                    val expression = builder.lower(joinAddress.get<PhysicalPostalAddressDb>("physicalPostalAddress").get("city"))
+                    val predicates = normalizedVariants.map { variant ->
+                        builder.like(expression, "%$variant%")
+                    }
+
+                    builder.or(*predicates.toTypedArray())
+                }
+            }
+
+        private fun byCountry(country: String?) =
+            Specification<LogisticAddressDb> { root, _, builder ->
+                country?.takeIf { it.isNotBlank() }?.let {
+                    builder.equal(root.get<LogisticAddressDb>("physicalPostalAddress").get<String>("country"),"%${it.lowercase()}%")
+                }
+            }
+
+        private fun normalizeGermanUmlauts(input: String): List<String> {
+            val base = input.lowercase();
+            return listOf(
+                base,
+                base.replace("ae", "ä"),
+                base.replace("oe", "ö"),
+                base.replace("ue", "ü"),
+                base.replace("ss", "ß")
+            ).distinct()
+        }
     }
 
     fun findByBpn(bpn: String): LogisticAddressDb?
--- a/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/BusinessPartnerSearchService.kt
+++ b/bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/BusinessPartnerSearchService.kt
@@ -22,9 +22,15 @@ package org.eclipse.tractusx.bpdm.pool.service
 import mu.KotlinLogging
 import org.eclipse.tractusx.bpdm.common.dto.PageDto
 import org.eclipse.tractusx.bpdm.common.dto.PaginationRequest
+import org.eclipse.tractusx.bpdm.pool.api.model.BusinessPartnerSearchFilterType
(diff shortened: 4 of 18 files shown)


nicoprow (commented):
@kunyao-cofinity-x 

A first feedback to the pull request:
There is now a lot of additional filters that can be combined with business partner searches. And these filters also include variant search clauses and and transforms like making everything lower case. With these very powerful search I could imagine that the performance might take a drastic hit on large data sets. Did you do a performance test
on your data set when using this search?

A possible more optimized solution that still only relies on the  database search capabilities could be to introduce search columns to relevant searchable columns. In the search columns you can then store the normalized text value of the value you want to search. By doing this you don't need to execute several variant clauses and don't need to transform text values on the fly.

kunyao-cofinity-x:
> @kunyao-cofinity-x
> 
> A first feedback to the pull request: There is now a lot of additional filters that can be combined with business partner searches. And these filters also include variant search clauses and and transforms like making everything lower case. With these very powerful search I could imagine that the performance might take a drastic hit on large data sets. Did you do a performance test on your data set when using this search?
> 
> A possible more optimized solution that still only relies on the database search capabilities could be to introduce search columns to relevant searchable columns. In the search columns, you can then store the normalized text value of the value you want to search. By doing this you don't need to execute several variant clauses and don't need to transform text values on the fly.

Hi @nicoprow, thanks for the feedback. For the performance improvement. I plan to add a few columns in the pool for the value normalization.
- legal_name_normalizing
- phy_street_name_normalizing
In those two columns, we would see the German umlaut. We can normalize them in advance so we don't convert them on the fly. Based on this idea, do you have a suggestion?

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/request/LegalEntityPropertiesSearchRequest.kt
```diff
@@ -26,9 +26,25 @@ import io.swagger.v3.oas.annotations.media.Schema
 @Schema(description = "Contains keywords used for searching in legal entity properties")
 data class LegalEntityPropertiesSearchRequest constructor(
     @field:Parameter(description = "Filter legal entities by name")
-    val legalName: String?
+    val legalName: String?,
+
+    @field:Parameter(description = "Filter business partners by CX-BPN.")
+    val id: String?,
```
nicoprow, line 32: Should match our naming convention: bpnL

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/request/LegalEntityPropertiesSearchRequest.kt
```diff
@@ -26,9 +26,25 @@ import io.swagger.v3.oas.annotations.media.Schema
...
+
+    @field:Parameter(description = "Filter business partners by street name.")
+    val street: String?,
```
nicoprow, line 35: Street is not a single field. We should talk about what is the expectation here if somebody puts something in that field. Should it search for partial matches in all street DB fields?

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/request/LegalEntityPropertiesSearchRequest.kt
```diff
@@ -26,9 +26,25 @@ import io.swagger.v3.oas.annotations.media.Schema
...
+
+    @field:Parameter(description = "Filter business partners by zip code.")
+    val postcode: String?,
+
+    @field:Parameter(description = "Filter business partners by city.")
+    val city: String?,
+
+    @field:Parameter(description = "Filter business partners by country code ISO 3166-1.")
+    val country: String?,
```
nicoprow, line 44: It says it should search by ISO country code but that would equal our enum value, so why is this a free text field? Or is it also expected to find matches on full country names like 'Germany'?

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/response/BusinessPartnerSearchResultDto.kt
```diff
@@ -0,0 +1,52 @@
...
+
+import io.swagger.v3.oas.annotations.media.Schema
+import org.eclipse.tractusx.bpdm.pool.api.model.LegalEntityIdentifierDto
+import org.eclipse.tractusx.bpdm.pool.api.model.LegalFormDto
+import org.eclipse.tractusx.bpdm.pool.api.model.StreetDto
+
+@Schema(name = "BusinessPartnerSearchResultDto", description = "")
+data class BusinessPartnerSearchResultDto(
+    @Schema(description = "BPN L/S/A")
+    val id: String,
```
nicoprow, line 30: Should match our naming convention: bpn

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/response/BusinessPartnerSearchResultDto.kt
```diff
@@ -0,0 +1,52 @@
...
+
+package org.eclipse.tractusx.bpdm.pool.api.model.response
+
+import io.swagger.v3.oas.annotations.media.Schema
+import org.eclipse.tractusx.bpdm.pool.api.model.LegalEntityIdentifierDto
+import org.eclipse.tractusx.bpdm.pool.api.model.LegalFormDto
+import org.eclipse.tractusx.bpdm.pool.api.model.StreetDto
+
+@Schema(name = "BusinessPartnerSearchResultDto", description = "")
+data class BusinessPartnerSearchResultDto(
```
nicoprow, line 28: We need to align about the requirements for such a generic business partner object. Currently, the Pool always distinguishes very strictly between legal entities, sites and addresses. Each of these business partner types has a distinct model. Does it make sense to have a 'generic' model that abstracts the information of each type like that?

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/response/BusinessPartnerSearchResultDto.kt
```diff
@@ -0,0 +1,52 @@
...
+    @Schema(description = "BPN L/S/A")
+    val id: String,
+
+    @Schema(description = "BPN L/S/A")
+    val name: String?,
+
+    @Schema(description = "Legal Form")
+    val legalForm: LegalFormDto?,
```
nicoprow, line 36: If this business partner is a site should it then have the legal form of its legal entity set in this field?

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/response/BusinessPartnerSearchResultDto.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    @Schema(description = "Street")
+    val street: StreetDto?,
+
+    @Schema(description = "City")
+    val city: String,
```
nicoprow, line 42: There are two possible cities for each business partner: the physical and the alternative city. Which city should be shown here in which cases?

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/response/BusinessPartnerSearchResultDto.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    @Schema(description = "PostalCode")
+    val postalCode: String,
+
+    @Schema(description = "Country")
+    val country: String,
+
+    @Schema(description = "")
+    val identifiers: List<LegalEntityIdentifierDto>
```
nicoprow, line 51: Here are only legal entity identifiers but if a business partner is an address it can also have address identifiers. What should be shown here in which cases?

nicoprow (commented):
I left some change requests and comments here. Some of these comments are actually regarding the underlying requirements. We can use those in the next refinement session as basis for debate

nicoprow:
Contributes to [#603](https://github.com/eclipse-tractusx/sig-release/issues/603)

kunyao-cofinity-x:
Two changes have been added

- Data model of search result. https://github.com/eclipse-tractusx/sig-release/issues/603#issuecomment-3158450186
- Normalized data columss have been added. https://github.com/eclipse-tractusx/bpdm/pull/1382#issuecomment-3061562502

kunyao-cofinity-x:
Hello @nicoprow @SujitMBRDI ,

Here is the summary of the commit:
- Search request "id" change to "bpn". It is allowed BPNL and BPNS.
- Search request "street" adapt StreetDto.
- Search parameters change from "only" to "Include", ex, IncludeLegalEntities, IncludeSites, IncludeAdditionalAddresses
- Search logic based on address (LogisticAddressRepository)
- No additional columns added, no additional maintenance efforts for the normalized value.
- Creating a normalized function in the database, normalized the value during the query of the table.
- Normalized function applies to legal name, street, and city.
- docs/api/pool.json and pool.yaml have been updated.

kunyao-cofinity-x:
Summary of the last commit:
- Refactoring the LogisticAddressRepository.searchBusinessPartner logic.
- Refactoring the BusinessPartnerSearchService.searchBusinessPartner logic.
- Fixing the issue when the street is null in the search request.

CC: @nicoprow @SujitMBRDI

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/request/LegalEntityPropertiesSearchRequest.kt
```diff
@@ -21,14 +21,30 @@ package org.eclipse.tractusx.bpdm.pool.api.model.request
...
 data class LegalEntityPropertiesSearchRequest(
     @field:Parameter(description = "Filter legal entities by name")
-    val legalName: String?
+    val legalName: String?,
+
+    @field:Parameter(description = "Filter business partners by CX-BPN.")
+    val bpn: String?,
+
+    @field:Parameter(description = "Filter business partners by street name.")
+    val street: StreetDto?,
```
nicoprow: Why is the whole StreetDto here? I can only see the street name field being used for the search. I recommend changing this field then to a string and rename it to streetname for clarity

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/BusinessPartnerSearchService.kt
```diff
@@ -199,4 +215,179 @@ class BusinessPartnerSearchService(
...
+        }
+
+        return PageDto(
+            totalElements = totalElements,
+            totalPages = totalPages,
+            page = page,
+            contentSize = size,
+            content = pagedContent
+        )
+    }
```
nicoprow: Something wrong about the pagination logic. If I have no data in my result I get a contentSize of 10 due to the pagination request size attribute.

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/repository/LogisticAddressRepository.kt
```diff
@@ -95,4 +98,60 @@ interface LogisticAddressRepository : JpaRepository<LogisticAddressDb, Long>, Jp
...
+                    OR (
+                            CAST(:#{#isAdditionalAddress} AS boolean) = true
+                            AND le.legal_address_id != la.id
+                            AND (s.id IS NULL OR s.main_address_id != la.id)
+                       )
+                )             
+            """,
+            nativeQuery = true
+    )
+    fun searchBusinessPartner(searchRequest: LegalEntityPropertiesSearchRequest, isLegalEntity: Boolean, isSite: Boolean, isAdditionalAddress: Boolean): Set<LogisticAddressDb>
```
nicoprow: The BPN search does not work for BPNAs, as the search only tries to match BPNLs and BPNSs. Is that intended?

nicoprow: Question For Business requirements: Does it make sens to specify a BPNA and return legal entities/sites who have a legal address/site main address matching that BPNA?

nicoprow: General notes for that query:

At the moment the query is error prone and will fail silently if the search request changes.

Also a normalizations are being executed on the fly for both columns and parameters. This is convenient but could lead to performance issues. We should keep a close eye on the performance for this feature and introduce optimizations if needed in the future.

Also, I recommend to convert this huge query to a JPA specifications to increase the maintainability of this database query.

nicoprow: Th query definitely needs to be paginated and return a Page instead of a set. Otherwise you will never be able to create a correct pagination. In order to do that you need to do two things:

1. Add a pageable as a parameter to the repository method
2. Set the return type to Page<LogisticAddressDb>

From my experience this should be enough for JPA to create a paginated query out of your definition. However, I never tried this with such complex queries. For some hintes have a look at this discussion: https://stackoverflow.com/questions/38349930/spring-data-and-native-query-with-pagination

The third answer is what I would go for first and only adapt if it somehow does not work.

bpdm-pool/src/test/kotlin/org/eclipse/tractusx/bpdm/pool/util/WebTestClientExtensions.kt
```diff
@@ -35,4 +35,3 @@ inline fun <reified T : Any> WebTestClient.invokeGetEndpoint(path: String, varar
         .responseBody
         .blockFirst()!!
 }
-
```
nicoprow: Unrelated to the feature. Please revert back for more clarity

nicoprow: I assume there must be some automatic optimization feature on your end then because it clearly shows that its different from the current state of main

nicoprow:
> Hi @nicoprow , thank you for the feedback. Here is the summary of the commit:
> 
>     1. Change street to streetName in the search request. The type is String.
> 
>     2. Return 400 and an error message when bpn, legalName, and the filter parameter are empty.
> 
>     3. Mapping alternativePostalAddress in the search result.
> 
>     4. Create DTOs under the Pool-API for pool search result, remove dependency of bpdm-gate-api
>        org.eclipse.tractusx.bpdm.pool.api.model.response.BusinessPartnerSearchResultDto
> 
>     5. revert org.eclipse.tractusx.bpdm.pool.service.compareStates, logic does not change, only indentation.
> 
>     6. Update CHANGELOG.md
> 
>     7. bpn search request support "BPNA"

Thank you for the adaptions. It looks good to me.

There are also other comments that are missing feedback. Will you work on them as well?

Regarding the mapping, the alternative address is now mapped correctly but the physical address still misses some fields (see the DTO).

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/MetadataService.kt
```diff
@@ -224,4 +224,4 @@ class MetadataService(
         val page = regionRepository.findAll(pageRequest)
         return page.toDto(page.content.map { it.toCountrySubdivisionDto() })
     }
-}
+}
```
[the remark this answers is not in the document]
kunyao-cofinity-x: This is reverted

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/controller/v6/LegalEntityLegacyServiceMapper.kt
```diff
@@ -554,7 +554,6 @@ class LegalEntityLegacyServiceMapper(
         legalEntity.currentness = createCurrentnessTimestamp()
 
         legalEntity.legalName = NameDb(value = legalEntityDto.legalName, shortName = legalEntityDto.legalShortName)
-
```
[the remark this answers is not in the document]
kunyao-cofinity-x: This is reverted.

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/repository/LogisticAddressRepository.kt
kunyao-cofinity-x: spport BPNA has been added.

kunyao-cofinity-x: Yes, I added OR la.bpn = :#{#searchRequest.bpn} to support bpna search

bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/BusinessPartnerSearchService.kt
```diff
@@ -199,4 +215,179 @@ class BusinessPartnerSearchService(
...
+
+    private fun shouldInclude(
+        searchResultFilter: Set<BusinessPartnerSearchFilterType>?,
+        filterType: BusinessPartnerSearchFilterType
+    ): Boolean {
+        return searchResultFilter.isNullOrEmpty() || searchResultFilter.contains(filterType)
+    }
+
+    private fun paginateResults(
```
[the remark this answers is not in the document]
kunyao-cofinity-x: The method is removed

bpdm-pool-api/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/api/model/request/LegalEntityPropertiesSearchRequest.kt
kunyao-cofinity-x: Adding a pattern validation in the services
searchRequest.country?.let {
            if (!it.matches(Regex("^[A-Z]{2}$"))) {
                throw BusinessPartnerSearchException("country must be ISO 3166-1 alpha-2 code, e.g. 'DE'")
            }
        }

nicoprow (approved):
LGTM

Issue #1390, opened 2025-07-02
Find Managed Legal Entities, Sites and Addresses and Their Relation to the Managing Legal Entity via the BPDM Pool Participant Endpoints


Sebastian-Wurm:
### Description
Use Case 1: As a Data Space Participant I want to query the BPDM Pool for the Managed Legal Entity and find the relation to the Managing Legal Entity, so that I can conclude a data exchange contract with the Managing Legal Entity about data of the Managed Legal Entity.

Use Case 2: As a Data Space Participant I want to query the BPDM Pool for sites and addresses of the Managed Legal Entity, so that I can use these in the Catena-X Use Cases.

(Future) Use Case 3: As a Data Space Participant and Managing Legal Entity I want to distinguish my own company data from the company data of my Managed Legal Entities, so that I can also manage legal entities, which are not belonging to my corporate group.

### Acceptance Criteria

- [x] Managed Legal Entities, their sites and address are treated as Data Space Participant business partner data records in the BPDM Pool (infer "Is Participant Data" in the BPDM Pool from "Is Managed By" relation additionally to "Is Own Company Data" from the BPDM Gate)
- [x] The "Is Managed By" relation to a Managing Legal Entity is returned together with the Managed Legal Entity

### Additional Information

Note that the current concept of indicating my own company data in the dedicated BPDM Gate, is to flag business partner data records, especially the single legal entity, as "Is Own Company Data". A future alternative concept of indicating the owner of a dedicated BPDM Gate could change how "Is Participant Data" is inferred. So this feature builds on the current concepts.

### Test Cases

Requirements: Have Identity of Managing Company

#### Legal Entity is Dataspace Participant If Shared By Its Managing Company

1. Share successfully own company legal entity representing the managing company (Should match the BPNL identity of the managing company)
2. Share successfully legal entity representing the managed company
3. Share successfully IsManagedBy relation between managing company and managed company (source: managed company, target: managing company)
4. Query the BPNL of the managed company in the Dataspace participants from the BPDM Pool
5. The managed company should appear in the query result

albireo411:
[...]

PR #1420, opened 2025-08-01
feat(bpdm-pool): enforce dataspace participant validation for managing entity in IsManagedBy relation

SujitMBRDI:
## Description

This pull request enhances the validation for IsManagedBy relation by condition that managing entity must be dataspace participant. Also, when relation validates successfully then made managed legal entity as dataspace participant as well.

Contributes to #1390  

## Pre-review checks


nicoprow (approved):
Tested the positive and the error case. Both works fine for me.

PR #1465, opened 2025-09-15
test(Pool): adapt tests to existing identifier types

nicoprow:
## Description

Tests now use the existing identifier types present in BPDM instead of using custom ones.

This pull request deletes the PoolDataHelper which was responsible for creating new identifier types before each test. Also the newer test data environment generator now sources from  the existing identifier types instead of creating own custom ones.

This is in line with how we also treat legal forms and administrative area level 1 regions. Therefore, making the test setup more consistent.

## Pre-review checks


nicoprow:
Changes already included in #1511

PR #1585, opened 2026-01-21
build(deps): increase spring boot starter to version 4.0.1

nicoprow:
## Description

This pull request increases the Spring Boot framework version all BPDM apps to 4.0.1.

Since this is a major version upgrade with a few breaking changes included it was necessary to adjust the BPDM code in various places. As a guideline I adhered to the migration guide for [Spring Boot version 4 from 3.x](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide).

A summary of all the necessary changes to make the new framework version work:

-  Replace Jackson 2.0 with Jackson 3.0. This means referencing a new Jackson dependency location and replacing the usage of the ObjectMapper with the now provided Spring configured JsonMapper
- Components of the actuator health indicator now switched packages and references needed to be adapted for all BPDM custom health check services
- Maven package locations have been deprecated for the  web starter, oauth2 resource server starter and oauth2 client starter. I replaced the old locations with all  the new locations
- oauth2 client registration components now also moved packaged names which affected all custom BPDM client configurations
- Support for the AntPathRequestMatcher for web security has been removed. I replaced it with directly using the the provided request matchers
- On various places more  Kotlin null checks where necessary now. This is due to the fact that the used methods/fields either became nullable I guess
- I removed the custom JSON deserializer we created for correctly deserializing Kotlin data class objects that have fields marked with @JsonUnwrapped annotation. Before, the deserialization of such fields was not possible due to a limitation of the jackson Kotlin module. Now however with Jackson 3.0 it seems to be possible out of the box. I removed the deserializer and all its annotations in the code
- JPA can now only combine non-null specifications. I adapted our specification factory methods to return an unrestricted specification (equivalent of a specification null-object) instead of null
- Flyway needs to be now integrated as a Spring Boot starter to be picked up during application startup
- The new hibernate version does not support creating empty embedded objects any longer. Before, hibernate offered a property  which when enabled, lets JPA create an empty embedded object even if all its values are null. We made use of that in the Orchestrator entities by having entities with non-null fields of embedded classes. Now, the only workaround I found for that is the use of a hibernate formular annotation to an useless  private field to enforce hibernate to actually create the object. This should be a temporary solution until we do a major refactoring to the Orchestrator entities
- For whatever reason code in the BPDM HTTP client proxies now need more type hints for Kotlin in order to work. Not sure, why that is now necessary but I added such type hints everywhere in the BPDM client definitions

## Pre-review checks


SujitMBRDI (approved):
LGTM

Issue #1646, opened 2026-04-30
feat: migrate Trivy workflow to reusable SHA-pinned action (TRG 8.04)


stephanbcbauer:
## Context

The current Trivy workflow(s) in this repository use **mutable tags** to reference `aquasecurity/trivy-action` (e.g. `@0.34.1`). Mutable tags can be silently redirected by a compromised upstream release, making the workflow a potential supply chain attack vector.

As part of the Tractus-X security hardening initiative, the Trivy workflow is being migrated to:

1. A **centralized reusable workflow** hosted in [eclipse-tractusx/sig-infra](https://github.com/eclipse-tractusx/sig-infra/blob/main/.github/workflows/reusable-trivy.yaml)
2. All action references inside it are **pinned to immutable commit SHAs**

This approach ensures the SHA is maintained in a single place and all consuming repositories automatically benefit from updates without individual patches.

## What needs to be done

Replace or extend the Trivy config/IaC scan with a lightweight caller that delegates to the reusable workflow:

```yaml
name: "Trivy"

on:
  schedule:
    - cron: "0 0 * * *"
  workflow_dispatch:

permissions:
  contents: read

jobs:
  trivy:
    uses: eclipse-tractusx/sig-infra/.github/workflows/reusable-trivy.yaml@main
    permissions:
      actions: read
      contents: read
      security-events: write
```

> **Note:** This repository runs a **docker image scan** with Trivy via a build matrix (`app-test-trivy.yaml`). That job is repo-specific and should remain local - ensure all `aquasecurity/trivy-action` references there are also **SHA-pinned**. The reusable workflow covers the config/IaC scan only.

## Acceptance Criteria

- [ ] Config/IaC Trivy scan calls the reusable workflow from `sig-infra`
- [ ] No mutable action tag references remain in any Trivy-related workflow file (including image scan workflows)
- [ ] Trivy scan results are still uploaded to the GitHub Security tab
- [ ] Workflow runs successfully on `main` branch after the change

## Related

- TRG 8.04: Mitigate high and above findings in Trivy
- Reusable workflow: [eclipse-tractusx/sig-infra - reusable-trivy.yaml](https://github.com/eclipse-tractusx/sig-infra/blob/main/.github/workflows/reusable-trivy.yaml)
- Tracking: eclipse-tractusx/sig-infra#567

tom-rm-meyer-ISST:
@nicoprow could you please review and check whether you want to use that?

[...]

PR #1649, opened 2026-05-08
cicd(workflows): pin trivy action version to specific SHA

nicoprow:
## Description

This pull request pins the trivy action that we use in our cicd pipeline to a specific SHA version, making sure that actual version used is not mutable.

Contributes to #1646

## Pre-review checks


nicoprow:
Trivy reports vulnerabilties but that means it works as expected. We should merge this pull request here to keep track of these reported vulnerabilities

PR #1704, opened 2026-06-01
feat(Pool): seed default reason code and remove upsert/delete endpoints

mohan06-mbrd:
## Description

Add default Reason Codes via Flyway migration.
Make Reason Codes read-only by removing update and delete APIs.

## Pre-review checks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,14 +18,7 @@ For changes to the BPDM Helm charts please consult the [changelog](charts/bpdm/C
 
 - Script variants and script codes for business partner data [#1593](https://github.com/eclipse-tractusx/bpdm/issues/1593)
 - BPDM Gate: Golden record relations to business partner output [#1630](https://github.com/eclipse-tractusx/bpdm/issues/1630)
-- ### Added
-
-- BPDM Pool: Seed default Business Partner Relation Reason Code via Flyway migration [#1679] (https://github.com/eclipse-tractusx/bpdm/issues/1679)
-
-### Changed
-
-- BPDM Pool: Make Reason Codes read-only by removing PUT and DELETE endpoints [#1687] (https://github.com/eclipse-tractusx/bpdm/issues/1687)
-
+- BPDM Pool: Seed default Business Partner Relation Reason Code via Flyway migration [#1679](https://github.com/eclipse-tractusx/bpdm/issues/1679)
 
 ### Changed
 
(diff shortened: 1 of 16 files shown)


mohan06-mbrd (commented):
Add default Reason Codes via Flyway migration.
Make Reason Codes read-only by removing update and delete APIs.

bpdm-pool/src/test/kotlin/org/eclipse/tractusx/bpdm/pool/v7/metadata/ReasonCodeGetV7IT.kt
```diff
@@ -20,56 +20,29 @@
...
+     * WHEN an operator requests reason codes
+     * THEN the default reason code HEADQUARTER_RELOCATION is present
      */
     @Test
-    fun `find created reason code`(){
-        //GIVEN
-        val request = ReasonCodeUpsertRequest(ReasonCodeDto(testName, "$testName description"))
-        poolClient.metadata.upsertReasonCode(request)
+    fun `default reason codes are seeded and returned by GET`() {
+        val expectedTechnicalKeys = setOf(
```
nicoprow: The expectetation should go into the THEN section according to our testing guideline. Also, should the expectation not be a pagination result?

see the original way of comparison

CHANGELOG.md
```diff
@@ -18,6 +18,14 @@ For changes to the BPDM Helm charts please consult the [changelog](charts/bpdm/C
 
 - Script variants and script codes for business partner data [#1593](https://github.com/eclipse-tractusx/bpdm/issues/1593)
 - BPDM Gate: Golden record relations to business partner output [#1630](https://github.com/eclipse-tractusx/bpdm/issues/1630)
+- ### Added
+
+- BPDM Pool: Seed default Business Partner Relation Reason Code via Flyway migration [#1679] (https://github.com/eclipse-tractusx/bpdm/issues/1679)
+
+### Changed
+
+- BPDM Pool: Make Reason Codes read-only by removing PUT and DELETE endpoints [#1687] (https://github.com/eclipse-tractusx/bpdm/issues/1687)
```
nicoprow: Actually the reason code endpoints have beeded created this release, so no need to add this changed entry here.  The changelog is for the end user being notified what has been done, we do not need to add changes here that revert things within one release as it will be not noticable to the end user.

nicoprow (changes requested):
Looks good to me, just needs some minor changes (see comments)

nicoprow (approved):
LGTM

Other runs failing is not because of this contribution:
1. Helm Upgrade is expected to fail because of breaking changes on the postgres dependencies
2. Trivy is reporting issues due to existing issues in the dependencies

PR #1736, opened 2026-07-22
Feat/ultimate owner isownedby hierarchy

mohan06-mbrd:
## Pre-review checks


bpdm-pool/src/main/kotlin/org/eclipse/tractusx/bpdm/pool/service/BusinessPartnerBuildService.kt
```diff
@@ -286,6 +286,10 @@ class BusinessPartnerBuildService(
             if (legalEntityBeforeUpdate != legalEntityAfterUpdate) {
```
nicoprow, line 286: The equivalence DTOs here also need to compare the ownershipUltimate flag. Otherwise, when you have an update that has ONLY the ownershipUltimate flag change, the validation below is skipped. Better add it the flag to the equivalence DTO. The additional check for whether we should validate the flag can stay like that.

nicoprow (changes requested):
A change needed to safeguard against a rare situation.

You first create parent and child linked by IsOwnedBy, each flagged via Pool's PUT /v7/legal-entities with an otherwise-identical payload:

  flag-parent=ok errors=[]   parent.ownershipUltimate=true  parent.ultimateOwnerBpnl=null
  flag-child =ok errors=[]   child.ownershipUltimate=true   child.ultimateOwnerBpnl=null

So in case you use the Pool API we could still have two flags at the same time. Solution sketched down below.

nicoprow (approved):
LGTM

PR #1782, opened 2026-08-10
Feature/ultimate owner system tests

mohan06-mbrd:
## Description

ultimate owner system tests

## Pre-review checks


nicoprow (changes requested):
The step definitions for the cucumber files do not exist in the system tester code and therefore the tests fail. The following step definitions are missing:

1. the {parent|grandparent|self-owner} entity {string} is marked as ownershipUltimate = true in the golden record
2. the golden record process processes the update
3. {string} output reflects ultimateOwnerBpnl as {string} / ... as null

Step 1 and 3 should be legitimitately implemented. Step 2 is unnecessary because the test helper methods already wait for the update to process. So all 'proccesses update' steps should be removed.

For executing the tests, the most straight forward way is to rebuild the system tester jar and then execute:

```bash
  java -jar bpdm-system-tester/target/bpdm-system-tester.jar \
    --threads 5 \
    classpath:cucumber/ultimate_owner_distribution.feature
```

You can also use intellij, find the system tester's Application.kt file, right click and then choose to run the Application.kt's main method. But that runs all tests and could result in a longer waiting time.

bpdm-system-tester/src/main/kotlin/org/eclipse/tractusx/bpdm/test/system/stepdefinations/UltimateOwnerDistributionStepDefs.kt
```diff
@@ -0,0 +1,139 @@
...
+            )
+        )
+        
+        // Update the legal entity in the Pool to mark it as ultimate owner
+        val updateRequest = LegalEntityPartnerUpdateRequest(
+            bpnl = legalEntityWithAddress.header.bpnl,
+            legalEntity = updatedLegalEntity
+        )
+        
+        poolClient.legalEntities.updateBusinessPartners(listOf(updateRequest))
```
nicoprow, line 116: This is not representative of the user flow. In a E2E scenario the sharing member designates an ultimate owner over the gate. This should be done here as well instead of taking an unrepresentative shortcut.

bpdm-system-tester/src/main/resources/cucumber/ultimate_owner_distribution.feature
```diff
@@ -0,0 +1,127 @@
...
+  ## An IsOwnedBy relation is established from the child to the parent.
+  ## The golden record process processes the updates.
+  ## The child entity reports the parent's BPNL as its ultimate owner.
+  @TEST_CXTPM-1743-001 @BPDM
+  Scenario: Owned entity reports ultimate owner from ownership chain
+    Given record "parent-record" reflects legal entity "parent"
+    And record "child-record" reflects legal entity "child"
+    When the sharing member shares relation "ownership" of type "IsOwnedBy" from "child-record" to "parent-record"
+    And the golden record process establishes relation "ownership"
+    And the parent entity "parent" is marked as ownershipUltimate = true in the golden record
```
nicoprow: This here is again two steps: The sharing member marks the legal entity as ultimate owner and the golden record process needs to confirm(establish) that the legal entity is indeeo the ultimate owner. Two steps, one for sharing member int gate, one for the golden record process service provider in the orchestrator.

For the naming there is a Gherkin convention that you try to talk 'business' here and not leak implementation details. Try not to use field names in the step definitions. Instead of 'ownershipUltimate = true' you can just use a description like 'marked as ultimate owner'.

bpdm-system-tester/src/main/resources/cucumber/ultimate_owner_distribution.feature
```diff
@@ -0,0 +1,127 @@
...
+  Scenario: Ultimate owner propagates through multi-level hierarchy
+    Given record "grandparent-record" reflects legal entity "grandparent"
+    And record "parent-record" reflects legal entity "parent"
+    And record "child-record" reflects legal entity "child"
+    When the sharing member shares relation "parent-ownership" of type "IsOwnedBy" from "parent-record" to "grandparent-record"
+    And the golden record process establishes relation "parent-ownership"
+    And the sharing member shares relation "child-ownership" of type "IsOwnedBy" from "child-record" to "parent-record"
+    And the golden record process establishes relation "child-ownership"
+    And the grandparent entity "grandparent" is marked as ownershipUltimate = true in the golden record
+    Then "parent-record" output reflects ultimateOwnerBpnl as "grandparent"
```
nicoprow: Also here, I would not use the field name but just basically write it out and it becomes 'business' talk: 'reflects "grandparent" as the ultimate owner'.

bpdm-system-tester/src/main/resources/cucumber/ultimate_owner_distribution.feature
```diff
@@ -0,0 +1,127 @@
...
+  ## IsOwnedBy relations are established between the entities.
+  ## The golden record process processes the hierarchy.
+  ## All entities report null as their ultimate owner.
+  @TEST_CXTPM-1743-003 @BPDM
+  Scenario: No ultimate owner reported when flag holder does not exist in chain
+    Given record "parent-record" reflects legal entity "parent"
+    And record "child-record" reflects legal entity "child"
+    When the sharing member shares relation "ownership" of type "IsOwnedBy" from "child-record" to "parent-record"
+    And the golden record process establishes relation "ownership"
+    Then "child-record" output reflects ultimateOwnerBpnl as null
```
nicoprow: reformulate to something more business here as well. 'has no ultimate owner' for example

nicoprow (changes requested):
The branch could use a merge from main. Other than that the test scope looks fine. The tests need more work in being more user flow representative. Also, step definitions could be more in 'business' talk.

nicoprow (approved):
LGTM

PR #1802, opened 2026-08-21
refactor: reorganize orchestrator service layer into application, par…

mohan06-mbrd:
…ser, and operation layers following pool pattern

## Description

refactor: reorganize orchestrator service layer..

## Pre-review checks


bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/application/GoldenRecordTaskApplicationService.kt
```diff
@@ -39,12 +42,12 @@ import java.time.Instant
 import java.util.*
 
 @Service
-class GoldenRecordTaskService(
+class GoldenRecordTaskApplicationService(
```
nicoprow, line 58: As an example, the application service here is the entry point for the controler (that is correct) but it contains actual operation service logic. Better to extract here validation logic to its own parser and the actual operation logic to its own operation service. This application service can then first validate the incoming request with the parser and then execute the operation if no errors have been detected.

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/operation/GoldenRecordTaskStateMachine.kt
```diff
@@ -34,7 +35,7 @@ import java.time.Instant
 class GoldenRecordTaskStateMachine(
```
nicoprow, line 35: Close to operation service already but should operate on a parsed model only. Also any kind of validations done here are better extracted to a parser. which is executed beforehand.

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/parser/GoldenRecordTaskRequestParser.kt
```diff
@@ -17,14 +17,14 @@
...
-package org.eclipse.tractusx.bpdm.orchestrator.service
+package org.eclipse.tractusx.bpdm.orchestrator.service.parser
 
 import org.eclipse.tractusx.bpdm.orchestrator.entity.*
 import org.eclipse.tractusx.orchestrator.api.model.*
 import org.springframework.stereotype.Service
 
 @Service
-class RequestMapper {
+class GoldenRecordTaskRequestParser {
```
nicoprow, line 27: This is more like a mapper than a parser. Even if you would want to make a parser out of  it here, the parser should first map from a unified request model to a parsed model and then actually return a parse result containing all the errors. A parser does not stop at the first encountered error and it never throws an exception for (user) invalid data.

nicoprow (changes requested):
Basic idea with the Orchestrator refactoring to the newest idiomatic design is to adhere to the [application code guide](https://github.com/eclipse-tractusx/bpdm/blob/main/docs/developer/application-code-guide.md).

Motivation:
1. We have different API models (V6, V7) but they contain a lot of overlap. We should make sure that business logic is reused between the different API versions where it makes sense
2. We have a big complex data (business partners) which contain a lot of possible validation errors. We should give the users a thorough validation result containing all possible errors in the response (instead of just giving the first encountered error)
3. We should aim for an agreed upon structure in the code so that the code becomes more understandable and more maintainable.

The code guide addresses these topics. The basic idea is the following:
1. We unify the different API models (V6, V7) into a common request model on which business logic can execute (validation and operation).
2. We handle incoming requests in a two-phase approach: First validating (parsing) the request completely and returning any user errors we can find. Secondly, if no errors exist we execute the actual operation on the validated request. 

In order to do this in an easy and structured way the code guide introduces the building blocks we can use:
1. Parsers: Validate, normalize and resolve unified user request data.
2. Operation Services: Execute the actual operation that was requested on the parsed user request (like create, update, search, or more complex operations)
3.  Application Services: Entry point for controllers. They bind parsers and operation services to create the full operation business logic (validation and exection).
4. Mappers: Should be employed when you need to reuse or extract translation code from one DTO model to another. The important part about mappers is that they only contain translation code and no business logic. 
5. DTO models: unified request model, parsed model, error model

Typically parsers and operation services are build in such a way that they should be reusable by each other. So when I use a parser for validation a business partner the parser makes use of cascading calls between lower level parsers until a full validation of the object is done. Operation services behave the same way.

Example for parser responsibilities: 
- Validate: Field content too long
- Normalization: "" -> null
- Resolution: recordId -> SharingMemberRecordDb

mohan06-mbrd:
Hi Nicko,
I have made changes to the Orchestrator refactoring. Please review them and let me know if any further changes are needed.
Thanks.

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/application/GoldenRecordTaskApplicationService.kt
```diff
@@ -39,45 +49,32 @@ import java.time.Instant
...
```
nicoprow, line 58: At the moment this is fine but once you convert more operations to the parser, operation structure the amount of services this application service is using will increase. Likely you will want to Split this service into operation scopes (GoldenRecordTaskCreateApplicationService, GoldenRecordTaskSearchApplicationService). Just a reminder to check for that, if this class grows too big to gets too many dependencies.

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/model/ParseResult.kt
```diff
@@ -0,0 +1,26 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ******************************************************************************/
+
+package org.eclipse.tractusx.bpdm.orchestrator.service.model
+
+sealed interface ParseResult<out T, out E> {
```
nicoprow, line 22: This one is a utlity class which is currently also used by the Pool. worth to extract a common type to the bpdm-commons and let Pool and Orchestrator both reference it.

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/model/GoldenRecordTaskCreateModels.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import org.eclipse.tractusx.orchestrator.api.model.BusinessPartner
+import org.eclipse.tractusx.orchestrator.api.model.TaskMode
+import java.util.UUID
+
+data class GoldenRecordTaskCreateRequest(
+    val index: Int,
+    val mode: TaskMode,
+    val recordId: String?,
+    val businessPartner: BusinessPartner
```
nicoprow, line 30: This DTO is actually a V7 DTO. Our current API model in the Orchestrator-API does not make that clear. There is V7 and a V6 BusinessPartner variant. Even though this is a big overhead I stell believe creating an API version independent BusinessPartner model has still benefit because currently the V6 version has to be awkwardly mapped to the V7 version. This increases cohesion between the API models. 

The cleaner way here is to create a API version independent BusinessPartner for the services.  Then you can map both the V7 and the V6 to this unified request model.

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/model/GoldenRecordTaskCreateModels.kt
```diff
@@ -0,0 +1,44 @@
...
+data class GoldenRecordTaskCreateRequest(
+    val index: Int,
+    val mode: TaskMode,
+    val recordId: String?,
+    val businessPartner: BusinessPartner
+)
+
+data class GoldenRecordTaskCreateParsed(
+    val mode: TaskMode,
+    val recordId: UUID?,
```
nicoprow: The parsing should resolve references. The unified request contains the recordId, therefore the parsed DTO should contain the resolved SharingMemberRecord entity.

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/operation/GoldenRecordTaskCreateOperation.kt
```diff
@@ -0,0 +1,66 @@
...
+@Transactional
+class GoldenRecordTaskCreateOperation(
+    private val stateMachine: GoldenRecordTaskStateMachine,
+    private val taskRepository: GoldenRecordTaskRepository,
+    private val sharingMemberRecordService: SharingMemberRecordApplicationService
+) {
+    private val logger = KotlinLogging.logger { }
+
+    fun create(parsed: List<GoldenRecordTaskCreateParsed>): List<GoldenRecordTaskDb> {
+        val gateRecords = sharingMemberRecordService.getOrCreateGateRecords(parsed.map { it.recordId })
```
nicoprow: This is not necessary anymore once the parser resolves these records beforehand

nicoprow (changes requested):
All in all this looks very good and shows a good understanding of the new code guideline.

I specifically checked the Golden Record Task Creation code as I saw this one was refactored. Code for the other operations I did not consider in my review and largely still not follow the code guideline (as expected).

Specific concerns in the comments below.

General concerns:

1. Having mapper and model package in the service package is not a bad idea but it currently does not match how it is structured in the Pool. I would propose to follow the Pool package structure for now (putting mapper and model in app root package) and we can discuss the package structure later.
2. Keep the scope of this pull request smaller. I propose to revert all services which are not correctly refactored to the code guideline back to their previous state. Keep this pull request only for the golden record task create operation. This way you can fully concentrate on this one slice.

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/model/GoldenRecordTaskCreateModels.kt
mohan06-mbrd: done

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/model/ParseResult.kt
mohan06-mbrd: done

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/application/GoldenRecordTaskApplicationService.kt
mohan06-mbrd: Noted. Will split into operation-scoped services when needed.

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/model/GoldenRecordTaskCreateModels.kt
nicoprow: But it is not done as far as I can see. This DTO still points to the V7 Orchestrator API DTO. The idea would be to have a API independent DTO which both V7 and V6 can map into.

mohan06-mbrd:
Noted. Hi Nicko, can we connect now for our daily sync if you are available?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

nicoprow:
outdated

PR #1821, opened 2026-08-28
Refactor/orchestrator/golden record task resolve

mohan06-mbrd:
## Pre-review checks


bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/parser/GoldenRecordTaskResolveParser.kt
```diff
@@ -0,0 +1,84 @@
...
+                    val task = foundTasks[resultEntry.taskId]
+                    when {
+                        task == null -> ParseResult.ofSingleFailure(GoldenRecordTaskResolveParseError.TaskNotFound(resultEntry.taskId))
+                        task.processingState.resultState == GoldenRecordTaskDb.ResultState.Aborted -> 
+                            ParseResult.ofSingleFailure(GoldenRecordTaskResolveParseError.TaskAborted(resultEntry.taskId))
+                        resultEntry.errors.isEmpty() -> {
+                            try {
+                                assertAdditionalSitesHaveSite(resultEntry.businessPartner)
+                                ParseResult.Success(GoldenRecordTaskResolveParsed(resolveRequest.step, task, resultEntry))
+                            } catch (e: BpdmInvalidBusinessPartnerException) {
```
nicoprow: I think it is unnecessary here to go for an assert and try catch block. better to just rewrite the assert to a non-throwing method. This code can be simplified this way

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/parser/GoldenRecordTaskResolveParser.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    private fun assertAdditionalSitesHaveSite(businessPartner: org.eclipse.tractusx.orchestrator.api.model.BusinessPartner) {
+        if (businessPartner.additionalSites.isNotEmpty() && businessPartner.site == null) {
+            throw BpdmInvalidBusinessPartnerException(
+                "additional sites of its address are stated but no site of its own is, which they would be additional to"
+            )
+        }
+    }
+
+    private fun toUuidOrNull(uuidString: String): UUID? =
```
nicoprow: Also here code duplication. Extract into util package for other services to reuse

nicoprow (changes requested):
Some minor code smells

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/parser/GoldenRecordTaskResolveParser.kt
mohan06-mbrd: Done! Changed from throw-based assertion to a validation method that returns an error message. Eliminated the unnecessary try-catch block for simpler code.

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/parser/GoldenRecordTaskResolveParser.kt
mohan06-mbrd: Done! Removed duplicate toUuidOrNull() method and now using the shared utility from the util package

nicoprow (approved):
Ok, this pull request looks good so far. I noticed that the toUUID function is duplicated along several services. I will merge this pull request here because there is already another pull request on top. So the deduplication can be done in the other pull request instead.

PR #1842, opened 2026-09-08
Refactor/orchestrator/batch and security services

mohan06-mbrd:
## Pre-review checks


bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/TimeoutProcessBatchService.kt
```diff
@@ -20,17 +20,16 @@
...
 import jakarta.annotation.PostConstruct
-import jakarta.persistence.EntityManager
 import mu.KotlinLogging
 import org.eclipse.tractusx.bpdm.orchestrator.config.TaskConfigProperties
+import org.eclipse.tractusx.bpdm.orchestrator.service.operation.TimeoutProcessBatchOperation
 import org.springframework.scheduling.annotation.Scheduled
 import org.springframework.stereotype.Service
 
 @Service
 class TimeoutProcessBatchService(
```
nicoprow: This service now acts as a configuration. Either rename and move to configuration package or remove completely and put the schedule annotation on the operartion service.

nicoprow (changes requested):
Ok, those two operations are quite different in what they are. The step security operation is used for securing endpoints. Actually, the code guideline at the moment has no strict place where to put those supporting services. For now I think your change is alright.

For the TimeoutProcessBatchService: this is a scheduled job. It is correct to put the actual service logic in an operation service. But then the TimeoutProcessBatchService should be removed. As it stands now it is a configuration merely just scheduling the timeout operation service. So either you move the logic to a proper configuration class or you just remove it completely and put the schedule annotation on top of the operation

Issue #1855, opened 2026-09-15
Orchestrator API: Refactpring Split Between V6 And V7


nicoprow:
## Motivation

The Orchestrator API (`bpdm-orchestrator-api`) does not separate V6 from V7. Some DTOs live in a
V6-only package, some are shared between both versions, and some API interfaces even declare V6 and
V7 request paths on the same method. V6 is frozen, so every shared type is a trap: a change intended
for V7 silently changes the frozen V6 contract, and nothing in the module structure or the compiler
prevents it.

Duplication is the accepted price here. Since V6 never changes again, a duplicated V6 type carries no
double-maintenance cost — it is written once and then left alone.

`bpdm-pool-api` already does this thoroughly and is the baseline to orientate against: the
unversioned root packages hold the current version (V7), everything V6 lives under `…/api/v6/**`
with a `V6` suffix on every type, and the only things V6 borrows from the root are `ApiCommons`
(path constants) and a single enum. There is no shared DTO and no shared API interface.

### 1. V6 DTOs reference V7 DTOs

`api/v6/model/BusinessPartner.kt` declares only `BusinessPartner` and `LegalEntity` (102 lines) and
wildcard-imports `org.eclipse.tractusx.orchestrator.api.model.*`. Every other type in the V6 payload —
`NamePart`, `UncategorizedProperties`, `Site`, `PostalAddress`, `BpnReference`, `Identifier`,
`BusinessState`, `ConfidenceCriteria`, `GoldenRecordType` — resolves to the V7 model in
`api/model/BusinessPartner.kt` (587 lines), which is the actively evolving one (script variants,
additional sites, golden record relations, ultimate owner, `updatedAt`). Any field added there today
appears in the frozen V6 request and response schemas tomorrow.

Further V6 → V7 model references:

| V6 file | V7 type it uses |
| --- | --- |
| `v6/model/TaskCreateRequest.kt` | `model.TaskMode` |
| `v6/model/TaskStepResultRequest.kt` | `model.TaskStep` |
| `v6/model/TaskStepResultEntryDto.kt` | `model.TaskErrorDto` (→ `TaskErrorType`) |
| `v6/model/TaskClientStateDto.kt` | `model.TaskProcessingStateDto` (→ `ResultState`, `StepState`, `TaskErrorDto`) |
| `v6/GoldenRecordTaskApi.kt` | `model.TaskStateRequest`, `model.TaskStepReservationRequest` |
[...]

PR #1856, opened 2026-09-16
Refactor orchestrator V6 API split

mohan06-mbrd:
## Pre-review checks


nicoprow:
Contributes to #1855

bpdm-orchestrator/src/main/kotlin/org/eclipse/tractusx/bpdm/orchestrator/service/ResponseMapper.kt
```diff
@@ -122,7 +122,11 @@ class ResponseMapper {
 
 
     fun toCategorizedNameParts(nameParts: List<NamePartDb>) =
-        nameParts.filter { it.type != null }.map { NamePart(it.name, it.type!!) }
+        nameParts.mapNotNull { namePart ->
+            namePart.type?.let { type ->
+                NamePart(namePart.name, type)
+            }
+        }
```
nicoprow: These changes have nothing to do with the split, right? Best take them out of the pull request

nicoprow (changes requested):
@mohan06-mbrd 
1. Currently GoldenRecordTaskCreateInboundMapperV6 maps V6 → V7 DTO → request model. But we should map directly from V6 to the existing request model.
2.  Enum conversions are duplicated 6+ times: TaskStepV6 ↔ TaskStep. Extract it to a mapper/v6 package mapper class.
3. You removed some V6 properties that were shipped already in a previous version. This is fine, as they came in accidentally when we changed the V7 model. However, we should note that in the CHANGELOG as a breaking change to the V6 model stating we fixed it by removing invalid changes from the previous release.
4. Update the openspec in the docs
5. unrelated changes in a different file

nicoprow (approved):
LGTM - but please also update the openAPI sepcs. Then we can merge

@mohan06-mbrd

nicoprow (changes requested):
Something is strange about the gate.yaml, how come there are so many changes here? To my understanding your changes only should have influenced the orchestrator.yaml and orchestrator.json api docs. please rollback the changes to the gate api docs and pool api docs

nicoprow (approved):
LGTM

